PR #3, opened 2023-11-09
2 migrate from setuppy to pyprojecttoml

elenbaasc:


pablolh:
I made some small changes to the project while you were working on it sorry :/ that was required to make the quantuminspire2 PR work...
So you're getting some merge conflicts which you'll have to fix.


No problem with the PR, except, why remove the `src` subfolder?
It's intention was to keep stuff internal to OpenSquirrel, so that people who `import opensquirrel` cannot use directly let's say SquirrelAST or other things which they shouldn't be using. It's done a lot in C++ but I heard the Python philosophy is different?
Does it make sense to have private implementation classes and an exposed API for a Python package? And if so how do you achieve it?

I'm really not an expert in Python imports and submodules...

pablolh:
It looks like what the privacy I mentioned is achieved with understore-prefixed class names:
https://stackoverflow.com/questions/551038/private-implementation-class-in-python

elenbaasc:
> I made some small changes to the project while you were working on it sorry :/ that was required to make the quantuminspire2 PR work... So you're getting some merge conflicts which you'll have to fix.
> 
> No problem with the PR, except, why remove the `src` subfolder? It's intention was to keep stuff internal to OpenSquirrel, so that people who `import opensquirrel` cannot use directly let's say SquirrelAST or other things which they shouldn't be using. It's done a lot in C++ but I heard the Python philosophy is different? Does it make sense to have private implementation classes and an exposed API for a Python package? And if so how do you achieve it?
> 
> I'm really not an expert in Python imports and submodules...

I'm also definitely not an expert. From my experience you would have a directory with the name of the package in the root directory that contains all the source files. I get your point of not giving access to certain modules. I believe they should then not be part of the package, but not sure how to arrange this properly. We should see how this is managed in Python in general.

opensquirrel/Circuit.py
```diff
@@ -1,42 +1,44 @@
...
+from opensquirrel.Replacer import Replacer
+from opensquirrel.SquirrelAST import SquirrelAST
+from opensquirrel.SquirrelASTCreator import SquirrelASTCreator
+from opensquirrel.SquirrelErrorHandler import SquirrelErrorHandler
+from opensquirrel.TestInterpreter import TestInterpreter
+from opensquirrel.TypeChecker import TypeChecker
+from opensquirrel.Writer import Writer
+from parsing.GeneratedParsingCode import CQasm3Lexer
+from parsing.GeneratedParsingCode import CQasm3Parser
+from test.TestGates import TEST_GATES as DefaultGates  # For the doctest.
```
pablolh, line 13: pretty sure this is now broken, since I moved test.TestGates to src.DefaultGates

elenbaasc: Yes, for some reason I didn't 'add' those files to the previous commit...

pablolh (approved):

PR #8, opened 2023-11-14
7 implement mkdocs for documentation

elenbaasc:


elenbaasc:
Not much changed to master, here. Just added mkdocs to pyproject.toml, some docs, and doc configuration.

mkdocs.yaml
```diff
@@ -0,0 +1,6 @@
+site_name: OpenSquirrel
+site_url: http://127.0.0.1:8000/
```
pablolh, line 2: This is localhost right? Is that intended?

pablolh (approved):

mkdocs.yaml
elenbaasc: Yes, this is intentional. The local host is also configured by default, but I left it such that we are reminded to update soon as we have a 'official' url.

PR #15, opened 2023-11-15
Set up GitHub actions

oschusler:


oschusler:
@elenbaasc this pipeline fails on two things;
1. In the `pyproject.toml` file you specify that the python version should be >= 3.10. Are you really sure you don't want to support 3.8 and 3.9? They are still in use
2. There is a rounding error on macOS. This should be fixed for the integration tests

pablolh:
Hi Olaf,

Thanks for the PR, I will fix the integration test and the Python versions

pyproject.toml
```diff
@@ -19,8 +19,6 @@ keywords = [
 classifiers = [
   'Development Status :: 3 - Alpha',
   'Programming Language :: Python :: 3',
-  'Programming Language :: Python :: 3.8',
-  'Programming Language :: Python :: 3.9',
```
elenbaasc, line 23: These should not be removed, right? As we want to still support 3.8 (as also described in the `pyproject.toml` file).

elenbaasc (approved):

pyproject.toml
pablolh: Oh sorry, I messed up
Will fix now

oschusler:
@pablolh the Quantum Device Connector is one of our internal applications and does not run client sided. This means that we can very much pin a specific python version and don't have to support all possible OS' and python versions.

oschusler:
> Ah! Numpy doesn't support Python 3.9 apparently. So it'll be 3.9+

There is no support in Numpy for < 3.10?

pablolh:
Sorry, I wrote it wrong. 
Numpy 1.25+ doesn't support 3.8, but does support 3.9:
see https://endoflife.date/numpy

We require Numpy 1.26 apparently. So alternatively I can accept Numpy 1.24

oschusler:
> Sorry, I wrote it wrong. Numpy 1.25+ doesn't support 3.8, but does support 3.9: see https://endoflife.date/numpy
> 
> We require Numpy 1.26 apparently. So alternatively I can accept Numpy 1.24

That is entirely up to you/the project. It only means that the classifiers and matrix in the actions need to be updated

pablolh:
I got it working!
I had to use multiple Python restricted dependencies in pyproject.toml:
using numpy 1.24 works with Python 3.8, but not Python 3.12
using numpy 1.26 works with 3.12 but not 3.8

I thought that the Numpy package would specify this and that Poetry would deal with this alone, but no, I was getting no clean error and some build failure. Not sure why, anyway, it works now.

Feel free to approve/merge whenever.

pyproject.toml
```diff
@@ -32,9 +32,12 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.10,<3.13"
+python = ">=3.8,<3.13"
 antlr4-python3-runtime = "^4.13.1"
-numpy = "^1.26.1"
+numpy = [
+    { version = "1.24", python = "<3.12" },
+    { version = "^1.26", python = "3.12" },
```
oschusler, line 39: Given that 1.26 works for everything 3.9 and later, wouldn't this be better?

```suggestion
    { version = "1.24", python = "<3.9" },
    { version = "^1.26", python >= "3.9" },
```

pyproject.toml
```diff
@@ -32,9 +32,12 @@ packages = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.10,<3.13"
+python = ">=3.8,<3.13"
```
oschusler, line 35: Normal notation is 
```suggestion
python = "^3.8"
```

pyproject.toml
pablolh: you are absolutely right it's fixed

pyproject.toml
pablolh: you are absolutely right it's fixed

PR #18, opened 2023-11-22
9 implement dev tools and document usage

elenbaasc:


opensquirrel/MatrixExpander.py
```diff
@@ -5,7 +5,7 @@
 
 
 # This should only be used for testing and on circuits with low number of qubits.
-def getBigMatrix(semantic, qubitOperands, totalQubits):
+def getBigMatrix(semantic: MultiQubitMatrixSemantic, qubitOperands, totalQubits):
```
pablolh, line 8: The correct type here is the parent class `Semantic` in the current code

pablolh: (see below `isinstance`)

pablolh (changes requested):

opensquirrel/MatrixExpander.py
elenbaasc: Yes, makes sense!

Perhaps we should make Semantic an ABC then?

pablolh: It is indeed an ABC! I just only recently discovered that you can use this in Python with an extra module - it's not in the language by default I believe.

pablolh (approved):

opensquirrel/MatrixExpander.py
elenbaasc: You need to import the `abc` library indeed (https://docs.python.org/3/library/abc.html).

I was also thinking if it would be an idea to create ABCs for the various passes. And Users can then define new passes based on those interfaces. We have done a similar thing with NetSquid.

PR #42, opened 2023-12-08
Optimize McKay decomposer

eendebakpt:
The main optimization is the replacement of `np.cross` with a direct calculation for 3d vectors. 

```
# on develop
%timeit myCircuit.decompose_mckay()
1.24 ms ± 54 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
# this PR
In [4]: %timeit myCircuit.decompose_mckay()
686 µs ± 79 µs per loop (mean ± std. dev. of 7 runs, 1,000 loops each)
```

@pablolh

### DIFF
--- a/opensquirrel/mckay_decomposer.py
+++ b/opensquirrel/mckay_decomposer.py
@@ -10,6 +10,7 @@ from opensquirrel.squirrel_ast import SquirrelAST
 
 FloatArray = npt.NDArray[np.float64]
 
+
 def normalizeAngle(x: float) -> float:
     t = x - 2 * pi * (x // (2 * pi) + 1)
     if t < -pi + ATOL:
@@ -18,10 +19,10 @@ def normalizeAngle(x: float) -> float:
         t -= 2 * pi
     return t
 
-def cross(l : FloatArray, r : FloatArray) -> FloatArray:
-    return np.array( [ l[1] * r[2] - l[2] * r[1],
-                      l[2] * r[0] - l[0] * r[2],
-                      l[0] * r[1] - l[1] * r[0] ], dtype=l.dtype) 
+
+def cross(l: FloatArray, r: FloatArray) -> FloatArray:
+    return np.array([l[1] * r[2] - l[2] * r[1], l[2] * r[0] - l[0] * r[2], l[0] * r[1] - l[1] * r[0]], dtype=l.dtype)
+
 
 class McKayDecomposer:
     def __init__(self, gates):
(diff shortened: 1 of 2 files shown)


opensquirrel/mckay_decomposer.py
```diff
@@ -66,7 +73,7 @@ def _flush(self, q):
 
     def _flush_all(self):
         while len(self.oneQubitGates) > 0:
-            self._flush(next(iter(self.oneQubitGates.keys())))
```
pablolh, line 69: nice, I didn't know

pablolh (approved):
I like your PR more for the nice code improvements than for the performance gain (we target feature and API over performance for now).
The replacement of `np.cross` for pure performance reasons is debatable to me - you add more code and the performance gain will likely not be noticeable (this algo takes roughly linear time in the number of gates).

eendebakpt:
> I like your PR more for the nice code improvements than for the performance gain (we target feature and API over performance for now). The replacement of `np.cross` for pure performance reasons is debatable to me - you add more code and the performance gain will likely not be noticeable (this algo takes roughly linear time in the number of gates).


Fair argument. Let me know if you want me to revert the changes to `np.cross`, I do not have a strong opinion here.

pablolh:
> Fair argument. Let me know if you want me to revert the changes to np.cross, I do not have a strong opinion here.

Let's revert the `np.cross` change then :)
I'll keep in mind that this `np.cross` seems to not be optimal performance-wise (for short vectors?).

eendebakpt:
In general numpy has a very large overhead for short arrays (or scalars). For example:
```
> import math
>  import numpy
> 
> %timeit math.sin(.2)
> %timeit numpy.sin(.2)
102 ns ± 8.93 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
710 ns ± 37.5 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```
And `np.cross` in particular has a lot of overhead because it is much more general than a simple 3d vector cross.

pablolh (approved):

eendebakpt:
> > In general numpy has a very large overhead for short arrays (or scalars).
> 
> That is very good to know indeed!
> 
> In this particular case I am using numpy for the mathematical functions it provides, not for its compact arrays and fast C implementation of linear algebra. So probably indeed not the best use case for numpy, since this is just a 3d vector... Do you know an alternative library for this? We could implement cross product with native Python types ourselves but I also feel this is already implemented somewhere... with tests and assertions and corner cases and...
> 
> (I am also using numpy in a more standard way when I build the circuit matrix, which is kind of circuit simulation, therefore performance-heavy and appropriate for numpy.)

Unfortunately, there is no good lightweight (e.g. performant for small arrays) alternative to numpy. There is https://github.com/kwant-project/tinyarray, but the arrays are immutable. One can use cython or numba to increase performance, or use extensions such as rust.

PR #47, opened 2023-12-13
38 fix types in the ast

pablolh:
Changes

- Implement ControlledSemantic to be able to express CNOT as a
  controlled X, for instance
- Separate validation of qubit ranges from type-checking in a separate
  ANTLR tree visitor
- snake_case instead of camelCase for variable names
- Add docstrings to parsing classes
- The GateLibrary superclass is used to factorize code for gateset-aware
classes (e.g. TypeChecker)
- Semantic classes are removed, absorbed into Gate subclasses
- Move ALL ANTLR parsing classes to parsing folder - this has already
  been done, but only partially
- Gates are specified directly with Python functions and a decorator
- Lay the basis of handling of anonymous gates
- Writer turned into impl pattern
- Rename AST -> IR
- Make test_testinterpreter a proper unit test, before it was testing
  parsing too
- Visitor pattern for the SquirrelIR class
- Use those visitors in e.g. writer and test_interpreter
- Various code fixes
- Use inspect to know Python function signatures
 This makes the gate semantic API nice to use from Python.


opensquirrel/default_gates.py
```diff
@@ -1,104 +1,74 @@
...
+    return ControlledGate(control, x(target))
+
+
+@named_gate
+def cz(control: Qubit, target: Qubit) -> Gate:
+    return ControlledGate(control, z(target))
+
+
+@named_gate
+def cr(control: Qubit, target: Qubit, theta: Float) -> Gate:
```
elenbaasc, line 67: `cr` by default is Controlled-Rz?

opensquirrel/mckay_decomposer.py
```diff
@@ -3,135 +3,115 @@
 
 import numpy as np
 
-from opensquirrel.common import ATOL, ArgType
-from opensquirrel.gates import SingleQubitAxisAngleSemantic, queryEntry, querySemantic, querySignature
-from opensquirrel.squirrel_ast import SquirrelAST
+from opensquirrel.common import ATOL, normalize_angle
+from opensquirrel.default_gates import rz, x90
+from opensquirrel.squirrel_ir import BlochSphereRotation, Float, Gate, Int, Qubit, SquirrelIR, Statement
```
elenbaasc, line 8:
```suggestion
from opensquirrel.squirrel_ir import BlochSphereRotation, Float, Gate, Qubit, SquirrelIR
```

opensquirrel/mckay_decomposer.py
```diff
@@ -3,135 +3,115 @@
...
+        combined_angle = 2 * acos(cos(a / 2) * cos(b / 2) - sin(a / 2) * sin(b / 2) * np.dot(l, m))
 
-        if abs(sin(combinedAngle / 2)) < ATOL:
-            self.oneQubitGates.pop(qubit)
+        if abs(sin(combined_angle / 2)) < ATOL:
+            self.accumulated_1q_gates.pop(qubit)
             return
 
-        combinedAxis = (
+        combined_axis = (
```
elenbaasc, line 81: PyCharm complains that this piece of code is unreachable. Is the above if statement always true?

opensquirrel/mckay_decomposer.py
```diff
@@ -3,135 +3,115 @@
...
-            self._processSingleOperation(operation)
+        if len(qubit_arguments) >= 2:
+            [self.flush(q) for q in qubit_arguments]
+            self.output.add_gate(gate)
+            return
 
-        self._flush_all()
+        if len(qubit_arguments) == 0:
+            assert False, "Unsupported"
+            return
```
elenbaasc, line 99: PyCharm complains that the `return` statement is unreachable, does the code exit when you have `assert False`?

opensquirrel/writer.py
```diff
@@ -1,41 +1,37 @@
-from opensquirrel.common import ArgType
-from opensquirrel.gates import querySignature
+from opensquirrel.squirrel_ir import Comment, Float, Gate, Int, Qubit, SquirrelIR, SquirrelIRVisitor, Statement
```
elenbaasc, line 1:
```suggestion
from opensquirrel.squirrel_ir import Comment, Float, Gate, Int, Qubit, SquirrelIR, SquirrelIRVisitor
```

opensquirrel/writer.py
```diff
@@ -1,41 +1,37 @@
...
-            return f"{squirrelAST.qubitRegisterName}[{arg}]"
-        if t == ArgType.INT:
-            return f"{int(arg)}"
-        if t == ArgType.FLOAT:
-            return f"{float(arg):.{Writer.NUMBER_OF_SIGNIFICANT_DIGITS}}"
+    def visit_qubit(self, qubit: Qubit):
+        return f"{self.qubit_register_name}[{qubit.index}]"
 
-        assert False, "Unknown argument type"
+    def visit_int(self, _int: Int):
```
elenbaasc, line 14: For consistency you could use `i` here as well, iso `_int`, as done in the `squirrel_ir.py` module. (same for `_float`).

opensquirrel/writer.py
```diff
@@ -1,41 +1,37 @@
...
+    def visit_float(self, _float: Float):
+        return f"{_float.value:.{self.number_of_significant_digits}}"
 
-        for operation in squirrelAST.operations:
-            if isinstance(operation, str):
-                comment = operation
-                assert "*/" not in comment, "Comment contains illegal characters"
+    def visit_gate(self, gate: Gate):
+        if gate.is_anonymous():
+            self.output += "<anonymous gate>\n"
```
elenbaasc, line 22: Add a hyphen?
```suggestion
            self.output += "<anonymous-gate>\n"
```

test/test_replacer.py
```diff
@@ -1,32 +1,31 @@
 import unittest
 
-from opensquirrel.default_gates import DefaultGates
-from opensquirrel.replacer import Replacer
-from opensquirrel.squirrel_ast import SquirrelAST
+from opensquirrel import replacer
+from opensquirrel.default_gates import *
+from opensquirrel.squirrel_ir import Gate, Qubit, SquirrelIR
```
elenbaasc, line 5:
```suggestion
from opensquirrel.squirrel_ir import Qubit, SquirrelIR
```

test/test_testinterpreter.py
elenbaasc, line 1: Rename in light of renaming the module that is being tested here.

elenbaasc (commented):
I have been through the code, but mostly to check if there are no weird things or things going wrong. I haven't been able to check/optimize the functionalities. It's quite a big refactor and a really nice one as well! I would still need some time to fully understand how everything links together, but our meeting this morning helped a lot.

The test coverage is now 82%, it would be good if we could increase this (perhaps this could be an issue of its own though).

pablolh:
Thanks Chris for the review, will look asap

opensquirrel/squirrel_ir.py
```diff
@@ -0,0 +1,225 @@
+import inspect
+from abc import ABC
+from dataclasses import dataclass
+from functools import wraps
+from typing import Callable, List, Optional, Tuple, Union
```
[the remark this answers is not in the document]
pablolh: Argh
I was hoping isort would do it for me

docs/reference.md
```diff
@@ -14,9 +14,9 @@
 
 ::: opensquirrel.replacer
 
-::: opensquirrel.squirrel_ast
+::: opensquirrel.squirrel_ir
 
-::: opensquirrel.squirrel_ast_creator
+::: opensquirrel.squirrel_ir_creator
```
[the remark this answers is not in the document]
pablolh: So we is the goal to not document things in parsing/ or utils?

docs/reference.md
```diff
@@ -14,9 +14,9 @@
```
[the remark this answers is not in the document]
pablolh: yes

pablolh:
I've tried increasing the test coverage a bit, by looking at what was not covered
Most of the non-covered code is now **generated code from ANTLR**... Not sure if I can do much about it, the `test_parsing.py` is already pretty complete

opensquirrel/squirrel_ir.py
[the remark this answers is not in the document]
elenbaasc: It is weird indeed that isort didn't pick this up....

docs/reference.md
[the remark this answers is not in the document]
elenbaasc: No, they should also be included in the docs.

elenbaasc:
> I've tried increasing the test coverage a bit, by looking at what was not covered Most of the non-covered code is now **generated code from ANTLR**... Not sure if I can do much about it, the `test_parsing.py` is already pretty complete

Yes, I see, good point. We could exclude the generated files by adding them to the `omit` list of the coverage tool in the `pyproject.toml`:
```toml
[tool.coverage.run]
branch = true
omit = ["tests/*", "**/.tox/**","opensquirrel/parsing/antlr/*"]
```

docs/reference.md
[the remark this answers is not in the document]
pablolh: I've automated this using a variant of this https://mkdocstrings.github.io/recipes/

scripts/gen_reference_page.py
```diff
@@ -0,0 +1,30 @@
...
+
+repo = Path(__file__).parent.parent
+opensquirrel = repo / "opensquirrel"
+reference_file = repo / "docs" / "reference.md"
+
+reference_md_content = "<!--- This file was automatically generated by scripts/gen_reference_page.py --->\n\n"
+for path in sorted(opensquirrel.rglob("*.py")):
+    module_path = path.relative_to(repo).with_suffix("")
+
+    print(module_path)
```
elenbaasc, line 17: Do we want to introduce `logging` for this, in stead of print statements? If so, we can add a TODO in the places and create an issue to pick this up later.

test/test_mckay_decomposer.py
```diff
@@ -0,0 +1,93 @@
+import unittest
+
+from opensquirrel import circuit_matrix_calculator, mckay_decomposer
+from opensquirrel.default_gates import *
+from opensquirrel.squirrel_ir import Float, Qubit, SquirrelIR
+
+
+def are_matrices_equal_up_to_global_phase(matrix_a, matrix_b):
```
elenbaasc, line 8:
```suggestion
def check_equivalence_up_to_global_phase(matrix_a, matrix_b):
```

test/test_mckay_decomposer.py
```diff
@@ -0,0 +1,93 @@
...
+    phase_difference = matrix_a[first_non_zero] / matrix_b[first_non_zero]
+
+    return np.allclose(matrix_a, phase_difference * matrix_b)
+
+
+class DecomposeMcKayTests(unittest.TestCase):
+    def checkMcKayDecomposition(self, squirrel_ir, expected_ir=None):
+        """
+        Check whether the mcKay decomposition transformation applied to the input IR preserves the
+        circuit matrix up to an irrelevant global phase factor.
```
elenbaasc, line 25:
```suggestion
        circuit matrix up to a global phase factor.
```

test/test_mckay_decomposer.py
```diff
@@ -0,0 +1,93 @@
...
+    if abs(matrix_b[first_non_zero]) < ATOL:
+        return False
+
+    phase_difference = matrix_a[first_non_zero] / matrix_b[first_non_zero]
+
+    return np.allclose(matrix_a, phase_difference * matrix_b)
+
+
+class DecomposeMcKayTests(unittest.TestCase):
+    def checkMcKayDecomposition(self, squirrel_ir, expected_ir=None):
```
elenbaasc, line 22:
```suggestion
    def check_mckay_decomposition(self, squirrel_ir, expected_ir=None):
```

opensquirrel/parsing/antlr/squirrel_ir_creator.py
```diff
@@ -0,0 +1,73 @@
...
+        GateLibrary.__init__(self, gate_set, gate_aliases)
+        self.squirrel_ir = None
+
+    def visitProg(self, ctx):
+        number_of_qubits, qubit_register_name = self.visit(ctx.qubitRegisterDeclaration())
+
+        self.squirrel_ir = SquirrelIR(number_of_qubits=number_of_qubits, qubit_register_name=qubit_register_name)
+
+        for gate_application in ctx.gateApplication():
+            self.visit(gate_application)
```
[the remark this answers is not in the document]
pablolh: I think `gate_application` is the right name, so I've fixed the two others

opensquirrel/squirrel_ir.py
```diff
@@ -0,0 +1,222 @@
...
+from dataclasses import dataclass
+from functools import wraps
+from typing import Callable, List, Optional, Tuple
+
+import numpy as np
+
+from opensquirrel.common import ATOL, normalize_angle
+
+
+class SquirrelIRVisitor(ABC):
```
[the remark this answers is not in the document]
pablolh: I guess this interface is just here to make the typing annotations work
It's equivalent to the base visitor class in tree-gen, for instance
A (somewhat complicated) explanation of the visitor pattern: https://en.wikipedia.org/wiki/Visitor_pattern

scripts/gen_reference_page.py
pablolh: This `print` was left accidentally, I didn't intend to leave it
now removed
Nevertheless, logging can be nice, so we can create the issue.

test/test_mckay_decomposer.py
pablolh: For me if I rename it that way, the function should return nothing and throw when matrices are not equal up to global phase
So
`check` -> assertion (function throws for instance)
`is` or `are` -> return a boolean, the `False` return value can also be useful in some context


I've renamed it as you suggest but then made it do the `self.assertTrue` on top and return void.

opensquirrel/squirrel_ir.py
```diff
@@ -0,0 +1,222 @@
...
+    def accept(self, visitor: SquirrelIRVisitor):
+        return visitor.visit_int(self)
+
+
+@dataclass
+class Qubit(Expression):
+    index: int
+
+    def __hash__(self):
+        return hash(self.index)
```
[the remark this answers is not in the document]
pablolh: I blindly followed this https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

In this case the qubit index alone would work perfectly I believe
But hashing functions can be tricky, some (pseudo)randomization (done by `hash()`) can be useful to get balanced hash tables, etc...

opensquirrel/squirrel_ir.py
```diff
@@ -0,0 +1,222 @@
...
+    generator: Optional[Callable[..., "BlochSphereRotation"]] = None
+
+    def __init__(self, qubit: Qubit, axis: Tuple[float, float, float], angle: float, phase: float = 0):
+        self.qubit: Qubit = qubit
+        self.axis = self._normalize_axis(np.array(axis).astype(np.float64))
+        self.angle = normalize_angle(angle)
+        self.phase = normalize_angle(phase)
+
+    @staticmethod
+    def _normalize_axis(axis):
```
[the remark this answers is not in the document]
pablolh: yes

opensquirrel/squirrel_ir.py
```diff
@@ -0,0 +1,222 @@
...
+class Gate(Statement, ABC):
+    arguments: Optional[Tuple[Expression, ...]] = None
+    generator: Optional[Callable[..., "Gate"]] = None
+
+    @property
+    def name(self) -> Optional[str]:
+        return self.generator.__name__ if self.generator else None
+
+    def is_anonymous(self) -> bool:
+        return self.arguments is None
```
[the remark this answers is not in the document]
pablolh: yes
tbh I'm not super fond of @property but it can be convenient sometimes

elenbaasc (approved):

PR #81, opened 2024-01-30
Some small changes to README.md.

rturrado:
Remove the manual of commands from the `MkDocs` section.
Say just how to build OpenSquirrel docs instead.


README.md
```diff
@@ -42,13 +44,11 @@ The documentation is generated using MkDocs. For full documentation visit [mkdoc
...
+In order to build `OpenSquirrel` documentation, run the following command in the root directory of `OpenSquirrel`:
 
-* `mkdocs new [dir-name]` - Create a new project.
-* `mkdocs serve` - Start the live-reloading docs server.
-* `mkdocs build` - Build the documentation site.
-* `mkdocs -h` - Print help message and exit.
-* `mkdocs gh-deploy` - Rebuild and deploy documentation to GitHub Pages.
+```shell
+mkdocs build
+```
```
elenbaasc, line 51: This is information for the developers of OpenSquirrel, they will need to know how to use MkDocs, any idea if there is another place we should keep this information? Makes sense not to have it in the README.md.

README.md
```diff
@@ -62,13 +62,13 @@ Project dependencies are specified in the `pyproject.toml` file in the root dire
 
 Website: <https://python-poetry.org/>
 
-Install dependencies
+Install dependencies.
```
elenbaasc, line 65:
```suggestion
Install dependencies:
```

README.md
```diff
@@ -62,13 +62,13 @@ Project dependencies are specified in the `pyproject.toml` file in the root dire
...
 
 ```shell
 $ poetry install
 ```
 
-Start a Poetry shell
+Start a Poetry shell.
```
elenbaasc, line 71:
```suggestion
Start a Poetry shell:
```

README.md
rturrado: Thanks for the comment, Chris.

I have a few personal `cheat sheet` guides that would inded better be in a share point, accessible to other engineers. For example, for `git`, for `docker`, for `OpenQL`/`libqasm`/`QX-simulator` builds, and so on. Typical operations you do once in a blue moon and you don't want to memorize. We could well have an `MkDocs` cheat sheet as well. What do you think?

Now, if for OpenSquirrel doc managing, you think that the developer would need to execute a few commands, I would, instead of just writing down a short manual, write a list of steps. E.g., for building OpenSquirrel docs, run `mkdocs new...`, then `mkdocs serve`, etc. I just assumed that an OpenSquirrel developer would only need to do the `build` step.

elenbaasc (dismissed):

README.md
rturrado: My point is:
- We could put it here, but I wouldn't just write a short MkDocs reference. I would instead write a set of steps that the OpenSquirrel developer may need to use, as I pointed out in my comment above.
- If that section was too large, we could link from here to a `docs/README.md`, where we could be more verbose because we would only talk about MkDocs there.

elenbaasc (approved):

PR #106, opened 2024-02-23
Added check for hybrid Control/Matrix Gate classes + test

juanboschero:


opensquirrel/squirrel_ir.py
```diff
@@ -7,10 +7,8 @@
 from typing import Callable, List, Optional, Tuple
 
 import numpy as np
-
```
rturrado, line 10: @elenbaasc, @pablolh Are we enforcing PEP styles somehow? I know CLion tell you if you need to add or remove newlines, I suppose PyCharm will do too. If @juanboschero is using VSCode, I understand you can use a plugin to do these checks for you.

opensquirrel/squirrel_ir.py
```diff
@@ -173,8 +170,10 @@ def __init__(self, matrix: np.ndarray, operands: List[Qubit], generator=None, ar
 
     def __eq__(self, other):
         # TODO: Determine whether we shall allow for a global phase difference here.
-        if not isinstance(other, MatrixGate):
+        if not isinstance(other, ControlledGate | MatrixGate):
             return False  # FIXME: a MatrixGate can hide a ControlledGate. https://github.com/QuTech-Delft/OpenSquirrel/issues/88
+        if isinstance(other,ControlledGate):
+            return _compare_gate_classes(self,other)
```
rturrado, line 176:
```suggestion
            return _compare_gate_classes(self, other)
```

opensquirrel/squirrel_ir.py
```diff
@@ -173,8 +170,10 @@ def __init__(self, matrix: np.ndarray, operands: List[Qubit], generator=None, ar
 
     def __eq__(self, other):
         # TODO: Determine whether we shall allow for a global phase difference here.
-        if not isinstance(other, MatrixGate):
+        if not isinstance(other, ControlledGate | MatrixGate):
```
rturrado, line 173: Should a `ControlledGate` inherit from `MatrixGate`? Like, all `ControlledGate`s are `MatrixGate`s. Just asking!

opensquirrel/squirrel_ir.py
```diff
@@ -200,8 +199,11 @@ def __init__(self, control_qubit: Qubit, target_gate: Gate, generator=None, argu
         self.target_gate = target_gate
 
     def __eq__(self, other):
-        if not isinstance(other, ControlledGate):
+       
+        if not isinstance(other, ControlledGate | MatrixGate):
```
rturrado, line 203: All these `ControlledGate | MatrixGate` checks... aren't a bit of "magic code"? Could we use an interface class of which these two would inherit, and that would express a given feature that both types of gate have? Again, if a `ControlledGate` were a `MatrixGate`, the check `ControlledGate |` wouldn't be needed.

test/utils/test_matrix_expander.py
```diff
@@ -75,3 +75,23 @@ def test_matrix_gate(self):
...
+                ]
+            ),
+            operands=[Qubit(1), Qubit(2)]
+        )
+
+        cnot_controlled_gate = ControlledGate(
+            Qubit(1), BlochSphereRotation(qubit=Qubit(2), axis=(1, 0, 0), angle=math.pi, phase=math.pi / 2)
+        )
+
+        self.assertTrue(cnot_controlled_gate == cnot_matrix_gate)
```
rturrado, line 97: This test looks really nice.

rturrado (commented):
Good work, @juanboschero !

Please consider my review as a set of comments and questions. I leave the final approval to Chris and Pablo.

Thanks!

opensquirrel/squirrel_ir.py
```diff
@@ -220,7 +222,23 @@ def get_qubit_operands(self) -> List[Qubit]:
...
 
+def _get_qubit_num(gate: ControlledGate | MatrixGate) -> int:
+    operand = gate.get_qubit_operands()
+    return max(qubit.index for qubit in operand) + 1
+
+def _compare_gate_classes(g1: ControlledGate | MatrixGate, g2: ControlledGate | MatrixGate) -> bool:
+    
+    from opensquirrel.utils.matrix_expander import get_matrix
+    
+    matrix_g1 = get_matrix(g1, _get_qubit_num(g1))
```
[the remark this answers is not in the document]
pablolh: A beginning of a new implementation could start by identifying and putting in separate variables the `MatrixGate` and the `ControlledGate`:

```python
assert type(g1) != type(g2)
matrix_gate = g1 if isinstance(g1, MatrixGate) else g2
controlled_gate = g1 if isinstance(g1, ControlledGate) else g2

# ...
```

opensquirrel/squirrel_ir.py
pablolh: I think you need to run `isort` and `black` to fix the code in order to pass static analysis in the CI

Agreed that my PyCharm tells me to use 2 blank lines between top level functions/classes

pablolh (changes requested):

pablolh:
Your implementation is correct but doesn't scale

I just had this idea, instead of my earlier comment:
You can still use the matrix I believe but you need to use the `_QubitReIndexer` from `replacer.py` (reindexer name courtesy of @elenbaasc)

opensquirrel/squirrel_ir.py
juanboschero: Yes, I am currently using VSCode with pylint checks however, for some reason, it recommends place a new line between the imports and class. I will need to check the PEP compliance I am using.

opensquirrel/squirrel_ir.py
juanboschero: Yes, this sounds a lot better. It could also open the door to importing controlled gates from other frameworks.

opensquirrel/squirrel_ir.py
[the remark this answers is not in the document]
juanboschero: Regarding your first comment, do you think it may be a good idea to break up the matrix into small square blocks of 2x2 and compare each block? I feel like this would be the safest way of performing a check without violating the memory constraints.

opensquirrel/squirrel_ir.py
rturrado: OK, thanks for the explanation Pablo. So we have here three disjoint set of gates then.

I understand this could be an implementation detail turn, but it appears to me that a ControlledGate would fit very well the role of _**decorator**_ of a gate (a modifier). So, the same as you have donuts and chocolate covered donuts, you would have single-qubit unitaries, e.g. Z, and controlled single-qubit unitaries, in this example controlled Zs.

.idea/.gitignore
pablolh, line 1: I'm not sure we want this folder in the repo

opensquirrel/replacer.py
```diff
@@ -43,10 +43,9 @@ def visit_controlled_gate(self, controlled_gate: ControlledGate):
         return result
 
 
-def get_replacement_matrix(replacement: List[Gate], qubit_mappings):
+def get_replacement_matrix(replacement: List[Gate], qubit_mappings: list[Qubit]):
```
pablolh, line 46: `get_replacement_matrix` is a name that doesn't match anymore with the uses of this function
- maybe move it to `matrix_expander.py`, along with the Reindexer
- maybe rename to `get_matrix_after_qubit_remapping` or a better name

opensquirrel/squirrel_ir.py
```diff
@@ -221,6 +225,20 @@ def is_identity(self) -> bool:
...
+def _compare_gate_classes(g1: ControlledGate | MatrixGate, g2: ControlledGate | MatrixGate) -> bool:
+
+    from opensquirrel.replacer import get_replacement_matrix
+
+    union_mapping = list(set(g1.get_qubit_operands()) | set(g2.get_qubit_operands()))
+
+    matrix_g1 = get_replacement_matrix([g1], union_mapping)
+    matrix_g2 = get_replacement_matrix([g2], union_mapping)
+    if matrix_g1.shape != matrix_g2.shape:
+        return False
```
pablolh, line 237: this `return False` will never be executed, because the mappings are identical
the matrix returned by `get_replacement_matrix` has shape completely determined by the `len` of the mapping

test/test_squirrel_ir.py
```diff
@@ -0,0 +1,82 @@
+import unittest
+
+import numpy as np
+
+from opensquirrel.default_gates import *
+from opensquirrel.utils.matrix_expander import get_matrix
```
pablolh, line 6: you don't use this and `utils/__init__.py`

test/test_squirrel_ir.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+        cnot_controlled_gate = ControlledGate(
+            Qubit(4), BlochSphereRotation(qubit=Qubit(100), axis=(1, 0, 0), angle=math.pi, phase=math.pi / 2)
+        )
+
+        self.assertTrue(cnot_controlled_gate == cnot_matrix_gate)
+
+    def test_cnot_inequality(self):
+        cnot_matrix_gate = MatrixGate(
```
pablolh, line 30: `swap_matrix_gate`?

test/test_squirrel_ir.py
```diff
@@ -0,0 +1,82 @@
...
+                [
+                    [1, 0, 0, 0],
+                    [0, 1, 0, 0],
+                    [0, 0, 1, 0],
+                    [0, 0, 0, 1],
+                ]
+            ),
+            operands=[Qubit(0), Qubit(2)],
+        )
+        small_cnot_gate = ControlledGate(
```
pablolh, line 60: wrong variable name?

opensquirrel/squirrel_ir.py
pablolh: I think you are right about the decorator, but it's already pretty much the case
This decorator still needs to inherit from `Gate` since it is a gate
And it needs to store a control qubit and a target gate, which is already done

Except if i'm missing something I think `ControlledGate` is already a "decorator" although we could make it more explicit by creating a `GateDecorator` interface from which `ControlledGate` would inherit

Then we can e.g. add the `PowerGate` decorator...

test/test_squirrel_ir.py
```diff
@@ -0,0 +1,82 @@
+import unittest
+
+import numpy as np
+
+from opensquirrel.default_gates import *
+
+
+class SquirrelIRTest(unittest.TestCase):
```
pablolh, line 8: bueno

.gitignore
```diff
@@ -153,6 +153,7 @@ dmypy.json
 cython_debug/
 
 # PyCharm
+.idea/
```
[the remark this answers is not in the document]
pablolh: Note: there might be cases when multiple commits per PR would be desired, when a PR does multiple orthogonal things.
However, when editing those multiple commits by addressing review comments, it requires a bit more git mastery, I do for instance "interactive git rebase" with either "fixup" commits or "breaks" to edit commits which are not the top of the stack.

This is worth learning as well, but the single commit option is easier.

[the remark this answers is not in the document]
pablolh: The point is that, while it is not an easy thing at times, the beauty of the git history (clear, minimalist, to the point) should be always considered, for your future self but also for your developer colleagues and possible power users/contributors.

One criterion that one can use, is that the code should be valid and all tests should pass when I go back in time at any commit of the whole history. This should be the case also when cutting between two commits of the same PR.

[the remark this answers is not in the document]
pablolh: To complete my train of thought: if you merge this PR, you will add 10 (!) commits to the repo, which seems really unnecessary given that you are only adding ONE feature requiring only limited overall code changes.

pablolh (dismissed):

opensquirrel/squirrel_ir.py
```diff
@@ -4,11 +4,19 @@
 from dataclasses import dataclass
 from functools import wraps
 from math import cos, sin
-from typing import Callable, List, Optional, Tuple
+from typing import Callable, List, Optional, Tuple, Union
```
rturrado, line 7: Is this `Union` still needed, used?

pablolh: No, it can be removed

pablolh:
Make sure you rebase on top of latest `origin/develop` before `push --force`

pablolh (approved):

PR #144, opened 2024-04-11
72 handle measure instructions and bit registers

juanboschero:


opensquirrel/merger.py
```diff
@@ -84,3 +85,6 @@ def merge_single_qubit_gates(squirrel_ir: SquirrelIR):
     for accumulated_bloch_sphere_rotation in accumulators_per_qubit.values():
         if not accumulated_bloch_sphere_rotation.is_identity():
             squirrel_ir.statements.append(accumulated_bloch_sphere_rotation)
+
+    squirrel_ir.statements = sorted(squirrel_ir.statements, key=lambda obj: isinstance(obj, Measure))
```
elenbaasc, line 89: What does this do? Does it do a resort such that the Measurements are placed at the end? Is the order of the other statements preserved (i.e. how they relate two any 2 qubit statements appearing)? Is this tested sufficiently?

opensquirrel/parsing/antlr/generated/CQasm3Parser.py
```diff
@@ -1030,6 +1030,8 @@ def exitRule(self, listener: ParseTreeListener):
 
         def accept(self, visitor: ParseTreeVisitor):
             if hasattr(visitor, "visitGateApplication"):
+                if "measure" in str(self.children[0]).lower():
```
elenbaasc, line 1033: We should not make any changes to any of the modules in the `opensquirrel/parsing/antlr/...` directory. We should remove all the calls to the antlr parser in the tests and replace them with the libqasm parser, where appropriate.

opensquirrel/replacer.py
```diff
@@ -41,6 +42,10 @@ def visit_controlled_gate(self, controlled_gate: ControlledGate):
         result = ControlledGate(control_qubit=control_qubit, target_gate=target_gate)
         return result
 
+    def visit_measurement_operation(self, m: Measure):
+        result = Measure(qubit=Qubit(self.mappings.index(m.qubit)), axis=m.axis)
+        return result
```
elenbaasc, line 47: The `_QubitReIndexer` only works on gates, so I'm not sure if this should be here. The `decompose` method skips anything other than `Gate`s, so it's not needed.

We could throw an error if the user wants to use the `Circuit.replace` method to replace a measure operation with a list of gates.

opensquirrel/squirrel_ir.py
```diff
@@ -77,10 +83,59 @@ def accept(self, visitor: SquirrelIRVisitor):
...
+    def __hash__(self):
+        return hash(self.index)
+
+    def __repr__(self):
+        return f"Bit[{self.index}]"
+
+    def accept(self, visitor: SquirrelIRVisitor):
+        return visitor.visit_bit(self)
+
+
```
elenbaasc, line 99: This is only required if we don't deprecate the antlr parser, which we will, so it's not needed.

opensquirrel/squirrel_ir.py
```diff
@@ -24,9 +24,15 @@ def visit_float(self, f: "Float"):
     def visit_qubit(self, qubit: "Qubit"):
         pass
 
+    def visit_bit(self, bit: "Bit"):
+        pass
+
```
elenbaasc, line 29: This is only required if we don't deprecate the antlr parser, which we will, so it's not needed.

opensquirrel/squirrel_ir.py
```diff
@@ -261,14 +339,24 @@ def accept(self, visitor: SquirrelIRVisitor):
...
+        self,
+        *,
+        number_of_qubits: int,
+        qubit_register_name: str = "q",
+    ):
         self.number_of_qubits: int = number_of_qubits
         self.statements: List[Statement] = []
         self.qubit_register_name: str = qubit_register_name
+        self.number_of_bits: int = number_of_qubits
+        self.bit_register_name: str = "b"
```
elenbaasc, line 352: No required yet.
```suggestion
```

opensquirrel/squirrel_ir.py
```diff
@@ -279,6 +367,12 @@ def __eq__(self, other):
         if self.qubit_register_name != other.qubit_register_name:
             return False
 
+        if self.number_of_bits != other.number_of_qubits:
+            return False
+
+        if self.bit_register_name != other.bit_register_name:
+            return False
+
```
elenbaasc, line 375: Not required yet
```suggestion
```

opensquirrel/writer.py
```diff
@@ -17,6 +17,10 @@ def visit_int(self, i: Int):
     def visit_float(self, f: Float):
         return f"{f.value:.{self.number_of_significant_digits}}"
 
+    def visit_measure(self, measurement: Measure):
+        formatted_args = (arg.accept(self) for arg in measurement.arguments)
+        self.output += f"{measurement.name} {', '.join(formatted_args)}\n"
```
elenbaasc, line 22: A measure operation can only have __one__ argument.
```suggestion
        self.output += f"{measurement.name} {measurement.arguments[0].accept(self)}\n"
```

test/test_integration.py
```diff
@@ -82,6 +82,59 @@ def test_simple(self):
...
+
+                qubit[3] qreg
+
+                Ry qreg[2], 2.34
+                Rz qreg[0], 1.5707963
+                Ry qreg[0], -0.2
+                CNOT qreg[1], qreg[0]
+                Rz qreg[0], 1.5789
+                CNOT qreg[1], qreg[0]
+                Rz qreg[1], 2.5707963
```
elenbaasc, line 98: This might be a bit overkill to only test the measure operation at the end.

We also want to test if the following measure operation works out correctly: 

```
measure qreg

//  should become

measure q[0]
measure q[1]
measure q[2]
```

opensquirrel/operation_library.py
```diff
@@ -0,0 +1,32 @@
+class OperationLibrary:
```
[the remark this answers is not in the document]
juanboschero: Do you think a good idea for now would be to rename this to InstructionLibrary for the time being?

I agree that we will need to revise how to differentiate instructions later on.

opensquirrel/merger.py
```diff
@@ -59,8 +59,8 @@ def merge_single_qubit_gates(squirrel_ir: SquirrelIR):
     while statement_index < len(squirrel_ir.statements):
         statement = squirrel_ir.statements[statement_index]
 
-        if not isinstance(statement, Gate):
-            # Skip, since statement is not a gate
+        if not isinstance(statement, Gate) and not isinstance(statement,Measure):
```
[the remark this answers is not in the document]
juanboschero: This can be done when the user inputs the code themselves. However, because of the decomposition step we add, the measurement statements get shifted up. Skipping immediately causes the logic of the decomposition to change. 

 i.e, this is what it looks like with the current implementation:
```
 qubit[3] qreg

Rz qreg[0], 1.5707963
X90 qreg[0]
Rz qreg[0], 2.9415927
X90 qreg[0]
Rz qreg[0], 3.1415927
CNOT qreg[1], qreg[0]
Rz qreg[0], -2.3521427
X90 qreg[0]
Rz qreg[0], 3.1415927
X90 qreg[0]
Rz qreg[0], 0.78945
CNOT qreg[1], qreg[0]
Rz qreg[2], 3.1415927
X90 qreg[2]
Rz qreg[2], 0.80159265
X90 qreg[2]
Rz qreg[1], -1.8561945
X90 qreg[1]
Rz qreg[1], 3.1415927
X90 qreg[1]
Rz qreg[1], 1.2853981
measure qreg[0]
measure qreg[2]
```

Adding the skip, changes it to this:
```
 qubit[3] qreg

Rz qreg[0], 1.5707963
X90 qreg[0]
Rz qreg[0], 2.9415927
X90 qreg[0]
Rz qreg[0], 3.1415927
CNOT qreg[1], qreg[0]
Rz qreg[0], -2.3521427
X90 qreg[0]
Rz qreg[0], 3.1415927
X90 qreg[0]
Rz qreg[0], 0.78945
CNOT qreg[1], qreg[0]
Rz qreg[1], -1.8561945
X90 qreg[1]
Rz qreg[1], 3.1415927
X90 qreg[1]
Rz qreg[1], 1.2853981
Rz qreg[2], 3.1415927
X90 qreg[2]
Rz qreg[2], 0.80159265
X90 qreg[2]
measure qreg[0]
measure qreg[2]
```
which does not match the other test cases for the decomposition

opensquirrel/merger.py
juanboschero: Yes, this is due to the decomposition step which puts them in the middle. I will create more tests.

opensquirrel/operation_library.py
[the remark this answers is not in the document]
elenbaasc: Sure, let's rename it so that it already aligns with the naming convention

opensquirrel/merger.py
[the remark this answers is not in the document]
elenbaasc: Oke we can leave it in for now, but since there is nothing wrong with the second output (i.e. it is semantically identical to the first output), we should change it at a later stage. I'll make an issue for it.

test/test_integration.py
juanboschero: I have added the test but not awaiting 0.6.3

opensquirrel/replacer.py
elenbaasc: The error (i.e. that a measure instruction cannot be replaced by a list of gates) shouldn't be part of a visitor method of the `_QubitReIndexer`, it should be part of the `replace` function in `replacer.py`.

test/test_integration.py
elenbaasc: I still miss a test where, e.g. 

```
qubit[3] qreg

H qreg[0]
CNOT qreg[0], qreg[1]
Z qreg[2]

measure qreg
```

becomes (not considering decomposition):

```
qubit[3] qreg

H qreg[0]
CNOT qreg[0], qreg[1]
Z qreg[2]

measure qreg[0]
measure qreg[1]
measure qreg[2]
```

We will release libqasm 0.6.3 soon and then be able to merge this PR.

elenbaasc:
Note that it is in general up to the reviewer to 'Resolve' a conversation, as they need to determine whether this is the case. If the owner of the PR accepts a code suggestion by the reviewer, then it can be considered to be resolved.

The `poetry.lock` files seem to still have some conflicts that need to be resolved before we can merge.

rturrado (commented):
Great work!

test/test_integration.py
juanboschero: I tested the output of three consecutive measurements with 0.6.3 i.e,

```
measure qreg[0]
measure qreg[1]
measure qreg[2]
```
and it has passed without error. 

Regarding, `measure qreg`, I think we create a separate issue for this

opensquirrel/parsing/antlr/squirrel_ir_from_string.py
```diff
@@ -33,7 +33,7 @@ def check_qubit_ranges_of_antlr_tree(tree):
     qubit_range_checker.visit(tree)  # FIXME: return error instead of throwing?
 
 
-def squirrel_ir_from_string(s: str, gate_set: list, gate_aliases: dict) -> SquirrelIR:
```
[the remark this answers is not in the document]
juanboschero: Quick heads up, I reverted the file to its original version.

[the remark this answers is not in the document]
elenbaasc: I'm working on this right now, will have it done in a bit :)

elenbaasc (approved):

PR #150, opened 2024-04-18
136 hectoqubit change gate angle parameters from radians to degrees

elenbaasc:


test/test_integration.py
```diff
@@ -333,25 +333,25 @@ def test_export_quantify_scheduler(self):
...
-                    "Rz(1.5707963, 'qreg[1]')",
-                    "Rxy(-0.19634954, 1.5707963, 'qreg[1]')",
-                    "Rz(-1.5707963, 'qreg[1]')",
-                    "CZ (qreg[0], qreg[1])",
-                    "Rz(0.19634954, 'qreg[0]')",
-                    "Rxy(-1.5707963, 1.5707963, 'qreg[0]')",
-                    "Rz(3.1415927, 'qreg[0]')",
-                    "Rz(3.1415927, 'qreg[1]')",
-                    "Rxy(1.5707963, 1.5707963, 'qreg[1]')",
+                    "Rz(-180, 'qreg[1]')",
```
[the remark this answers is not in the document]
rturrado: Oh sorry, I read `180` and thought that `pi` would be clearer. Especially `pi/16` for `11.25`. But that wouldn't be `pi`, you're right. Maybe `180/16` instead of `11.25` would be clearer? The rest of the angles are quite self descriptive.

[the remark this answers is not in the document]
elenbaasc: This is a test to check the generated output of the quantify scheduler transpiler. I don't see value in changing the transpiler such that it provides the output in fractions (also not sure how Quantify scheduler deals with fractions: they might be rounded); this will, in general, not be read by humans.

test/test_integration.py
juanboschero, line 1: If it is important, maybe a test where the output is written in degrees is a nice to have.

juanboschero (commented):
It looks good. The only issues I see is if we also want to output the CQASM file in degrees

test/test_integration.py
juanboschero: The warning could be triggered by setting values in between 2pi and 10. Also, we could incorporate some sort of optional flag or kwargs that is None by default but the user can set it manually to indicate degrees vs. radians i.e using_degrees: False.

juanboschero: For the time being, I will resolve the changes, but it may be interesting to open a ticket regarding this.

juanboschero (approved):
Looks good.

PR #175, opened 2024-05-01
Fix syntax parameterized gates.

elenbaasc:


opensquirrel/exporter/writer.py
```diff
@@ -25,8 +25,12 @@ def visit_gate(self, gate: Gate):
             self.output += "<anonymous-gate>\n"
             return
 
-        formatted_args = (arg.accept(self) for arg in gate.arguments)
-        self.output += f"{gate.name} {', '.join(formatted_args)}\n"
+        gate_name = gate.name
+        if any(not isinstance(arg, Qubit) for arg in gate.arguments):
```
rturrado, line 29: This is my only doubt for this review. If `not isinstance(arg, Qubit)` will be the best check in the future. E.g., if we will ever have `Bit`, `int`, `bool` types... even function calls like `2*f() -> float`.  In that case, it would be better to be more explicit: e.g., if `arg` can be interpreted as a float.

elenbaasc: I agree, it is not very future proof. Note that a parameter for a parameterized gate, currently, may be either `Float` or `Int`, so that's why I cannot just focus on it being a `Float`. What would be better is that the parameters and qubit arguments are separated already at another stage (e.g. in the SquirrelIR), now they are all part of the function arguments.

This will very likely need to be changed as soon as we make changes to the IR. Including a refactor of the Writer pass in general.

Do you think we should change this bit of code before merging, or is it ok for now?

rturrado: As you wish. The could would look like `isinstance(arg, (Int, Float))`, which may be a bit more informative. Anyway, this wouldn't be yet future proof if we ever add `Bool`. :)

elenbaasc: I would rather keep it as is. As any argument, that is not a qubit, is considered a parameter (for now).

rturrado (approved):

PR #176, opened 2024-05-01
Initial register management implementation

rturrado:
### Add `RegisterManager` class
A circuit has a register manager.
A register manager only holds, for the moment, a (virtual) qubit and bit register size and names.
The bit register size is set to that of the qubit register.

### Add `Mapping` class
A mapping is a dictionary from a (virtual) qubit register to a physical qubit register.
Mapping correctness is a class invariant. I.e., it is checked during the construction of the mapping.

### Change `relabel` code into `_QubitRemapper`
`_QubitRemapper` implements `SquirrelIRVisitor` to update the qubit indices according to a `Mapping`.
 
### Update `circuit.map`
Remove `maq_qubits` free function and rename `circuit.map_qubits` to `circuit.map` (to be consistent with `circuit.decompose` and so on).
A mapper is created with a valid mapping and then returns its mapping through `get_mapping()`.
`circuit.map` invokes `remap_ir` with a given `mapper.get_mapping()`, and `remap_ir`  uses an `SquirrelIRVisitor` to update the qubit indices.
 
### Remove `SquirrelIR`'s number of qubits and qubit register name
This information is accessed through the circuit's register manager.

### Move `_QubitReindexer` code to `reindexer/qubit_reindexer.py`.
Rename `check_valid_replacement` to `check_gate_replacement`.
Rename `get_matrix_after_qubit_remapping` to `get_circuit_matrix(get_reindexed_circuit())`.
Rename `qubit_mappings` to `qubit_indices`.

### Rename `parser/libqasm/libqasm_ir_creator.py` to `parser/libqasm/parser.py`.


juanboschero (dismissed):

juanboschero (approved):

opensquirrel/mapper/mapping.py
```diff
@@ -24,5 +24,17 @@ def __eq__(self, other):
...
+    def __len__(self) -> int:
+        return len(self.data)
+
     def size(self) -> int:
         return len(self.data)
+
+    def items(self) -> List[Tuple[int, int]]:
+        return self.data.items()
+
+    def keys(self) -> List[int]:
```
juanboschero, line 36: Is the list of keys supposed to output a list of integers?

opensquirrel/squirrel_ir.py
```diff
@@ -114,7 +114,7 @@ def __eq__(self, other):
         return self.qubit == other.qubit and np.allclose(self.axis, other.axis, atol=ATOL)
 
     def accept(self, visitor: SquirrelIRVisitor):
```
juanboschero, line 116: This should have a return type

juanboschero (changes requested):

opensquirrel/squirrel_ir.py
rturrado: Yup. I still have to update my branch with the changes from Stan's PR (the type hinting one).

test/reindexer/test_qubit_reindexer.py
```diff
@@ -0,0 +1,55 @@
+from typing import List
```
juanboschero, line 1: For completeness, it may be wise to add a measurement test after reindexing.

juanboschero (changes requested):

test/reindexer/test_qubit_reindexer.py
rturrado: Thanks! I've changed the `replacement_gates_4` input so that it uses all the visited methods in `_QubitReindexer`.

Issue #89, opened 2024-02-05
Optimize the McKay decomposition


pablolh:
The McKay decomposer is not optimal in some corner cases

For instance:

It turns x into [zm90, x90, x90, zm90] which I didn't even know was correct (but it is)
doing [x90, x90] would be enough

PR #195, opened 2024-05-21
Implement naïve McKay decomposition optimization and match anonymous gates to defined gates

juanboschero:
This also covers issue #89


test/test_integration.py
```diff
@@ -355,6 +355,29 @@ def test_export_quantify_scheduler(self):
...
+        myCircuit = Circuit.from_string(
+            """
+            version 3.0
+
+            qubit[1] q   // Qubit (register) declaration
+
+            Y90 q[0]
+            X q[0]
+
+            """
```
rturrado, line 368:
```suggestion
            """
```

opensquirrel/squirrel_ir.py
```diff
@@ -148,6 +148,14 @@ def name(self) -> Optional[str]:
 
     @property
     def is_anonymous(self) -> bool:
```
rturrado, line 150: Maybe add a doc for this function? It's quite readable, but I think it would help. 

So, to check if a gate is anonymous, we walk all the default gate set, filter those which have one argument, assume that that one argument is a qubit (maybe worth adding a check/assert for this?), check that the axis and phase of that qubit argument for the default gate is the same as ours, and, in that case, convert ourselves to that default gate?

But there's something I don't like about this code. It has lateral effects. The functions says `is_anonymous`, but it should be called something like `convert_me_to_a_default_gate_if_possible`. I suggest this is done through different functions. E.g.:
- `is_anonymous` just checks if the gate is anonymous or not. I don't know how useful this function can be.
- `is_close_to_a_default_gate` could return `None` if it's  not, or `Gate` if it is (with the code below).
- `try_name_anonymous_gate(s)` could be a free function that goes through one/many anonymous gates and, if they are close to a default gate, name them (convert  them).

rturrado (changes requested):
Great work Juan!

Please review the `is_anonymous` function and its usage throughout the code.

We should really avoid having functions that:
- have lateral effects,
- don't do what the name suggests (e.g., `is_anonymous` changes the object),
- do multiple things (i.e., the name would be something like `does_this_and_that`),
- ...

In the end, it all comes to choosing good function names. They are usually clear and beautiful.

rturrado:
Also, please check these static analysis errors. I think you'll be safe running the commands below and `git push`ing the suggested changes:

```
(opensquirrel-py3.11) OpenSquirrel $ poetry run isort .
(opensquirrel-py3.11) OpenSquirrel $ poetry run black .
(opensquirrel-py3.11) OpenSquirrel $ python -m pytest -vv
```

opensquirrel/squirrel_ir.py
juanboschero: I have now added a _check_if_anonymous function that does the check. I do believe it should return true or false. This function is a bit strange because it accepts a `BlochSphereRotation`, not a `Gate`.

rturrado (changes requested):

opensquirrel/decomposer/mckay_decomposer.py
```diff
@@ -23,6 +23,9 @@ def decompose(g: Gate) -> [Gate]:
         if abs(g.angle) < ATOL:
             return []
 
+        if g.name == "Rz" or g.name == "X90":
```
eendebakpt, line 26: Even though this decomposition is more efficient for many backends (including the qi backends), it is not McKay decomposition. If kept this way, then the docstring should be updated.

opensquirrel/squirrel_ir.py
```diff
@@ -137,6 +137,25 @@ def __init__(self, generator, arguments):
...
+        if self.arguments is not None:
+            return False
+        from opensquirrel.default_gates import default_gate_set
+
+        for _, gate in enumerate(default_gate_set):
+            gate_args = inspect.signature(gate).parameters.values()
+            if len(list(gate_args)) == 1:
+                output_bloch = gate(self.get_qubit_operands())
+                if np.allclose(output_bloch.axis, self.axis) and np.allclose(output_bloch.phase, self.phase):
+                    Gate.__init__(self, gate, self.get_qubit_operands())
```
eendebakpt, line 155: Does this line have an effect?

rturrado: Even if it did, I find it obscure calling a constructor from a method. I see this should be reserved for calling a base class constructor from a derived class constructor.

Following up on my comment above, we would consider this code suggestion below.

    class Gate()
        def _set_qubit_operands(self, qubit_operands):
            self.qubit_operands = qubit_operands

        def _find_close_default_bloch():
        """Search a default BlochSphereRotation that is close to this gate.
        
        Returns:
            A default BlockSphereRotation if there is any close to this gate, or None otherwise.
        """
        # current _check_is_anonymous implementation,
        # but returning a BlochSphereRotation (true case) or None (false case)

        def is_anonymous():
            return self.arguments is None
    
        def try_name_as_default_bloch():
            default_bloch = self._find_close_default_bloch(self)
            if default_bloch is not None:
                self._set_qubit_operands(default_block.get_qubit_operands())

    # Somewhere in the project
    if gate.is_anonymous():
        gate.try_name_as_default_bloch()

test/test_integration.py
```diff
@@ -234,34 +225,29 @@ def test_qi(self):
...
-
         expected = """version 3.0
 
 qubit[4] q
 
 X90 q[1]
 Rz(1.5707963) q[1]
 X90 q[1]
-Rz(-0.2) q[0]
+Rz(-0.20000005) q[0]
```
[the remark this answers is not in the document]
juanboschero: This is the case for the two other comments

rturrado (approved):
All fine with me. Great  work, Juan!

PR #198, opened 2024-05-22
Rename opensquirrel to open_squirrel

rturrado:
Rename `opensquirrel` to `open_squirrel`.
Rename `SquirrelIR` to `IR`.
Rename `SquirrelIRVisitor` to `IRVisitor`.
Rename `squirrel_ir` to `ir`.

I am keeping `Statement`, `Gate`, `Measure`… classes within `ir.py`. That means that, for the time being, they still have to be imported from `open_squirrel.ir`. I don’t think that is such a bad thing (definitely better than `opensquirrel.squirrel_ir`). And, since they are classes that actually belong to the IR structure, I would leave them like that. However, I leave this up to debate, since it goes against one of the acceptance criteria for this task.


open_squirrel/utils/identity_filter.py
```diff
@@ -1,4 +1,4 @@
-from opensquirrel.squirrel_ir import Gate
+from open_squirrel.ir import Gate
 
 
 def filter_out_identities(l: [Gate]):
```
juanboschero, line 4: I don't know if we still need an identity filter, after our chat with Richard, he mentioned he did want explicit identities. 

However, this needs be in place when we decompose.

test/test_circuit_builder.py
```diff
@@ -19,12 +19,24 @@ def test_simple(self):
...
         assert circuit.qubit_register_size == 2
         assert circuit.qubit_register_name == "q"
-        assert circuit.squirrel_ir.statements == [
+        assert circuit.ir.statements == [
             Comment("A single line comment."),
             H(Qubit(0)),
             CNOT(Qubit(0), Qubit(1)),
         ]
 
+    def test_identity(self):
```
juanboschero, line 28: We should also add one to the integration, i.e test to see if the identity gate can be read on cqasm and incorporated on the IR

juanboschero (approved):

test/test_circuit_builder.py
rturrado: I have updated `test_libqasm.py`, `test_simple`, adding a `I q[0]` line. That is basically doing the cQASM v3 parsing and creating an IR, then comparing it with another circuit built through the circuit builder.

Let me know if you think that's enough. I can add a specific integration test otherwise, or update an existing integration test.

open_squirrel/utils/identity_filter.py
rturrado: No idea about this. Does this have to do with the Identity gate being added to OpenSquirrel?

PR #205, opened 2024-05-29
181-get_default_bloch

rturrado:
Main changes:
- Change `merge_single_qubit_gates` so that it checks if a BSR is a default BSR instead of an anonymous BSR before appending it (`merger/general_merger.py` ).
- Fix `BlochSphereRotation.get_default_bloch` so that it also checks for angle closeness. Notice `BlochSphereRotation.get_default_bloch` is basically the code we had in `Gate._check_is_anonymous` (`ir.py`).
- Add `default_bloch_sphere_rotations_without_params`, and `default_bloch_sphere_rotations` sublists to `default_gate_set` (`default_gates.py`).

Minor changes:
- Revert changes to `Gate.is_anonymous` (`ir.py`).
- Change `Gate.name` to print `<anonymous-gate>` (`ir.py`).
- Change `visit_gate` so it doesn't hardcode `<anonymous-gate>` for an anonymous gate (`exporter/writer.py`).


opensquirrel/ir.py
```diff
@@ -247,14 +210,25 @@ def is_identity(self) -> bool:
         # Angle and phase are already normalized.
         return abs(self.angle) < ATOL and abs(self.phase) < ATOL
 
-    def relabel(self, mapping: Mapping[int, int]) -> None:
-        """Relabel the qubits using the given mapping.
+    def get_default_bloch(self) -> BlochSphereRotation | None:
```
juanboschero, line 213: Hey Roberto, great pull request! Well done!

It may be interesting to name this function `wrap_default_bloch` and return a `BlochSphereRotation` only. Thus, if the function does not find a default bloch, it returns itself.

This may be better in the future and also help clear some lines in the general_merger.

opensquirrel/merger/general_merger.py
```diff
@@ -75,18 +80,19 @@ def merge_single_qubit_gates(squirrel_ir: SquirrelIR):
...
                 statement_index += 1
         statement_index += 1
 
     for accumulated_bloch_sphere_rotation in accumulators_per_qubit.values():
         if not accumulated_bloch_sphere_rotation.is_identity():
-            squirrel_ir.statements.append(accumulated_bloch_sphere_rotation)
-
-    squirrel_ir.statements = sorted(squirrel_ir.statements, key=lambda obj: isinstance(obj, Measure))
+            default_bloch = accumulated_bloch_sphere_rotation.get_default_bloch()
+            if default_bloch is not None:
```
juanboschero, line 96: If a `BlochSphereRotation` is returned (no None), you can delete this check.

juanboschero (approved):

opensquirrel/ir.py
rturrado: Good idea, yes, I prefer that than calling the get function and then check against none.

But then let's implement this as a free function receiving a BSR, instead of a BSR method, something like `try_name_anonymous_bloch(bsr: BSR) -> BSR`. That receives a BSR and returns the same BSR if it doesn't find a close default BSR, or a close default BSR otherwise.

Because we're not actually wrapping anything,  we're trying to convert an anonymous bloch into a default bloch.

We can put this free function in the merger,  for the time being, and invoke it as `blah = try_name_anonymous_bloch(blah)`.

opensquirrel/merger/general_merger.py
rturrado: Beautiful, yes!

PR #222, opened 2024-06-10
Remove annotation check at runtime

### LINKED ISSUE #219: Type annotations at runtime
The `named_gate` and `named_measruement` wrappers check the type annotations at runtime. This is undesirable as it is
1) Not Pythonic.
2) Dangerous, there are edge cases were type annotations can be correct by an error is raised (`"opensquirrel.ir.Qubit"` would be a correct typehint for an `Expression`, but an error will be raised.)

These checks should be removed and handled by a static type checker like `mypy`.

S-Linde:
Remove annotation check at runtime for the `named_gate` and `named_measurement` wrappers.

Closes #219


rturrado (approved):
Good for me, Stan!

Thanks!

PR #223, opened 2024-06-11
208 axis object

S-Linde:
This merge request contains the following:
- An `Axis` object to handle axis parsing, validation and operations.
- An `AxisLike` type alias to annotate what can be converted to this `Axis` object.

One of the core ideas of the `Axis` object is that it can take any input that looks like an axis and create a axis from it. Hence it improves the user friendliness of out program. For example `Axis((1,2,3))`, `Axis(1,2,3)`, `Axis(1.0, 2.0, 3.0), `Axis(np.array([1,2,3]))` and `Axis([[1],[2],[3]])` are all allowed and result in equal axis objects.

During initialization and setting of the `Axis.axis` property, the underlying 1DArray is normalized. Hence, an `Axis` object is *always* normalized.

### DIFF
--- a/opensquirrel/ir.py
+++ b/opensquirrel/ir.py
@@ -5,7 +5,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Callable
 from dataclasses import dataclass
 from functools import wraps
-from typing import Any, Sequence, TypeAlias, overload
+from typing import Any, Sequence, TypeAlias, cast, overload
 
 import numpy as np
 from numpy.typing import ArrayLike, NDArray
@@ -87,8 +87,9 @@ class Qubit(Expression):
 
 class Axis(Sequence[np.float64], Expression):
 
-    def __init__(self, axis: AxisLike) -> None:
-        self._axis = self._parse_axislike(axis)
+    def __init__(self, *axis: AxisLike) -> None:
+        axis_to_parse = axis[0] if len(axis) == 1 else cast(AxisLike, axis)
+        self._axis = self._parse_axislike(axis_to_parse)
 
     @property
     def axis(self) -> NDArray[np.float64]:
@@ -115,7 +116,7 @@ class Axis(Sequence[np.float64], Expression):
         return normalize_axis(axis)
 
     def __getitem__(self, index: int, /) -> np.float64:  # type:ignore[override]
-        return self.axis[index]  # type: ignore[no-any-return]
+        return cast(np.float64, self.axis[index])
 
     def __len__(self) -> int:
         return 3
--- a/opensquirrel/ir.py
+++ b/opensquirrel/ir.py
@@ -8,9 +8,9 @@ from functools import wraps
 from typing import Any, Sequence, TypeAlias, cast, overload
 
 import numpy as np
-from numpy.typing import ArrayLike, NDArray
+from numpy.typing import ArrayLike, DTypeLike, NDArray
 
-from opensquirrel.common import ATOL, are_matrices_equivalent_up_to_global_phase, normalize_angle, normalize_axis
+from opensquirrel.common import ATOL, are_matrices_equivalent_up_to_global_phase, normalize_angle
 
 
 class IRVisitor(ABC):
@@ -99,8 +99,8 @@ class Axis(Sequence[np.float64], Expression):
     def axis(self, axis: AxisLike) -> None:
         self._axis = self._parse_axislike(axis)
 
-    @staticmethod
-    def _parse_axislike(axis: AxisLike) -> NDArray[np.float64]:
+    @classmethod
+    def _parse_axislike(cls, axis: AxisLike) -> NDArray[np.float64]:
         if isinstance(axis, Axis):
             return axis.axis
 
@@ -113,7 +113,11 @@ class Axis(Sequence[np.float64], Expression):
             raise ValueError(
                 f"Axis requires an ArrayLike of length 3, but received an ArrayLike of length {len(axis)}."
             )
-        return normalize_axis(axis)
+        return cls._normalize_axis(axis)
+
+    @staticmethod
+    def _normalize_axis(axis: NDArray[np.float64]) -> NDArray[np.float64]:
+        return axis / np.linalg.norm(axis)
 
     def __getitem__(self, index: int, /) -> np.float64:  # type:ignore[override]
         return cast(np.float64, self.axis[index])
@@ -137,14 +141,14 @@ class Measure(Statement, ABC):
     def __init__(
         self,
         qubit: Qubit,
-        axis: ArrayLike = (0, 0, 1),
+        axis: AxisLike = (0, 0, 1),
         generator: Callable[..., Measure] | None = None,
         arguments: tuple[Expression, ...] | None = None,
     ) -> None:
         self.generator = generator
         self.arguments = arguments
         self.qubit: Qubit = qubit
-        self.axis = normalize_axis(np.array(axis).astype(np.float64))
+        self.axis = Axis(axis)
 
     def __repr__(self) -> str:
         return f"Measure({self.qubit}, axis={self.axis})"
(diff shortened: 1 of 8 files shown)


test/test_ir.py
```diff
@@ -1,16 +1,81 @@
...
 import pytest
+from numpy.typing import ArrayLike
 
 from opensquirrel.common import ATOL
-from opensquirrel.ir import BlochSphereRotation, ControlledGate, MatrixGate, Measure, Qubit
+from opensquirrel.ir import Axis, BlochSphereRotation, ControlledGate, Expression, MatrixGate, Measure, Qubit
+
+
+class TestAxis:
+
```
rturrado, line 16:
```suggestion
```

opensquirrel/ir.py
```diff
@@ -82,6 +85,103 @@ def accept(self, visitor: IRVisitor) -> Any:
...
+        """
+        return axis / np.linalg.norm(axis)
+
+    def __getitem__(self, index: int, /) -> np.float64:  # type:ignore[override]
+        """Get the item at `index`."""
+        return cast(np.float64, self.axis[index])
+
+    def __len__(self) -> int:
+        """Length of the axis, which is always 3."""
+        return 3
```
rturrado, line 161: What do you think of defining 3 (magic number) as a `_len` class member? And then `return _len` from here.

opensquirrel/ir.py
```diff
@@ -387,3 +487,7 @@ def __repr__(self) -> str:
     def accept(self, visitor: IRVisitor) -> None:
         for statement in self.statements:
             statement.accept(visitor)
+
+
+# Type Aliases
+AxisLike = Union[ArrayLike, Axis]
```
rturrado, line 493: Just a comment, a double check. Not sure how portable across Python versions this `Union` is.

opensquirrel/utils/matrix_expander.py
```diff
@@ -145,6 +146,21 @@ def visit_matrix_gate(self, gate: MatrixGate) -> NDArray[np.complex_]:
         return expanded_matrix
 
 
+X = np.array([[0, 1], [1, 0]])
+Y = np.array([[0, -1j], [1j, 0]])
+Z = np.array([[1, 0], [0, -1]])
```
rturrado, line 151:
```suggestion
x_axis = np.array([[0, 1], [1, 0]])
y_axis = np.array([[0, -1j], [1j, 0]])
z_axis = np.array([[1, 0], [0, -1]])
```

Trying to avoid variable names starting with uppercase if possible, although I admit `X`, `Y`, and `Z` are much more readable. Just a suggestion.

test/test_ir.py
```diff
@@ -1,16 +1,81 @@
...
+    @pytest.mark.parametrize("expected_class", [Sequence, Expression])
+    def test_inheritance(self, axis: Axis, expected_class: type[Any]) -> None:
+        assert isinstance(axis, expected_class)
+
+    def test_axis_getter(self, axis: Axis) -> None:
+        np.testing.assert_array_equal(axis.axis, [1, 0, 0])
+
+    @pytest.mark.parametrize(
+        "new_axis, expected_axis", [([0, 0, 1], [0, 0, 1]), ([0, 3, 4], [0, 3 / 5, 4 / 5]), (Axis(0, 1, 0), [0, 1, 0])]
```
rturrado, line 29:
```suggestion
        "new_axis, expected_axis", [
            ([0, 0, 1], [0, 0, 1]),
            ([0, 3, 4], [0, 3 / 5, 4 / 5]),
            (Axis(0, 1, 0), [0, 1, 0])
        ]
```

Just a suggestion, for readability, a few lines following the pattern `(new_axis, expected_axis)`.

test/test_ir.py
```diff
@@ -1,16 +1,81 @@
...
+    def test_get_item(self, axis: Axis) -> None:
+        assert axis[0] == 1
+        assert axis[1] == 0
+        assert axis[2] == 0
+
+    def test_len(self, axis: Axis) -> None:
+        assert len(axis) == 3
+
+    def test_repr(self, axis: Axis) -> None:
+        assert repr(axis) == "Axis[1. 0. 0.]"
```
rturrado, line 63: Hm... why was I expecting `Axis[1., 0., 0.]`? But I have to say I prefer this repr instead, yes.

test/test_ir.py
```diff
@@ -162,8 +227,8 @@ def test_inequality(self, gate: BlochSphereRotation, other_gate: BlochSphereRota
     def test_get_qubit_operands(self, gate: BlochSphereRotation) -> None:
         assert gate.get_qubit_operands() == [Qubit(42)]
 
-    def test_is_identity(self, gate: BlochSphereRotation) -> bool:
-        assert BlochSphereRotation.identity(42).is_identity()
+    def test_is_identity(self, gate: BlochSphereRotation) -> None:
+        assert BlochSphereRotation.identity(Qubit(42)).is_identity()
```
rturrado, line 231: The Answer to the Ultimate Question of Life, the Universe, and Everything! 😀

rturrado (commented):
Impressive, @S-Linde!

I'm sorry my Python knowledge is not of great help here.

I hope to be of use though to make you check a few points here and there.

Thanks!

opensquirrel/ir.py
```diff
@@ -199,9 +299,9 @@ def __eq__(self, other: object) -> bool:
         if abs(self.phase - other.phase) > ATOL:
             return False
 
-        if np.allclose(self.axis, other.axis):
+        if np.allclose(self.axis.axis, other.axis.axis):
```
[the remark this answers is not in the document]
S-Linde: I agree, done.

opensquirrel/ir.py
S-Linde: Done

opensquirrel/ir.py
S-Linde: Type aliases are a hot mess, currently these are the preferred ways to do it based on python version:
Python312 (not yet compatible with `mypy`):
```python
type AxisLike = ArrayLike | Axis
```
Python 310 and 311 (compatible with `mypy`):
```python
from typing import TypeAlias

AxisLike: TypeAlias = ArrayLike | Axis
```
Python 38 and 39 (compatible with `mypy`):
```python
from typing import Union

AxisLike = Union[ArrayLike, Axis]
```

Usually I would take the newest version and hide it behind an ```if TYPE_CHECKING:``` block. However, we need this alias in other modules as well, were we then would also need such blocks. Instead I chose the the python 38/39 version of type aliasing as it currently works for all python versions we support. When we drop support for python 39 (Okt-2025) we should update this accordingly.

opensquirrel/utils/matrix_expander.py
S-Linde: I would prefer to conform to the pep8 style guide as much as possible, which says that constants should be `UPPER_CASE_WITH_UNDERSCORES` ([see](https://peps.python.org/pep-0008/#constants))

test/test_ir.py
S-Linde: I agree, done

test/test_ir.py
S-Linde: This is the default representation of a `numpy` array, it was already used somewhere else, so I thought it would be nice to use it here as well.

test/test_ir.py
S-Linde: 😄

opensquirrel/ir.py
rturrado: Thanks for the detailed explanation.

Maybe worth opening an issue to log a future needed change? Something like `Update type aliases when we drop support for Python 3.9 (October 2025)`?

opensquirrel/utils/matrix_expander.py
rturrado: OK, beautiful!

test/test_ir.py
rturrado: Fair enough!

opensquirrel/ir.py
S-Linde: Made issues for the end of life of python3.8 (#227) and python3.9 (#228)

rturrado: Thanks!

rturrado (approved):

PR #229, opened 2024-06-13
224 fix angle ambiguity and measure statement in quantify scheduler exporter

elenbaasc:


opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -29,19 +32,32 @@ def __init__(self, qubit_register_name: str):
...
+        qubit_index = g.qubit.index
+        self.schedule.add(
+            quantify_scheduler_gates.Measure(
+                self._get_qubit_string(g.qubit),
+                acq_channel=qubit_index,
+                acq_index=qubit_index,
+                acq_protocol="ThresholdedAcquisition",
+            )
+        )
+        return
```
rturrado, line 44:
```suggestion
        self.schedule.add(
            quantify_scheduler_gates.Measure(
                self._get_qubit_string(g.qubit),
                acq_channel=g.qubit.index,
                acq_index=g.qubit.index,
                acq_protocol="ThresholdedAcquisition",
            )
        )
        return
```

I think `g.qubit.index` is quite self explanatory to not needing to define a `qubit_index`.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -19,8 +19,11 @@
     "Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
 )
 
+DEG_PRECISION = 5
```
rturrado, line 22: Maybe add a comment here (doc comment)? Something like: `Quantify-scheduler expected output precision`.

test/test_integration.py
```diff
@@ -421,12 +422,25 @@ def test_export_quantify_scheduler() -> None:
...
+            operation.data["gate_info"]
+            for operation in exported_schedule.operations.values()
+            if operation.data["gate_info"]["operation_type"] == "measure"
+        ]
+
+        for i, ir_measurement in enumerate(ir_measurements):
+            qubit_index = ir_measurement.qubit.index
+            assert qs_measurements[i]["acq_channel_override"] == qubit_index
+            assert qs_measurements[i]["acq_index"] == qubit_index
+            assert qs_measurements[i]["acq_protocol"] == "ThresholdedAcquisition"
```
rturrado, line 442: I would probably wouldn't use a `qubit_index` var here either. But it's OK with me to leave it.

test/exporter/test_quantify_scheduler_exporter.py
```diff
@@ -63,12 +64,16 @@ def test_export(self):
                 [
                     unittest.mock.call(theta=FloatEq(math.degrees(math.pi)), phi=FloatEq(0), qubit="q[0]"),
                     unittest.mock.call(
-                        theta=FloatEq(math.degrees(1.23)), phi=FloatEq(math.degrees(math.pi / 2)), qubit="q[2]"
+                        theta=FloatEq(round(math.degrees(1.23), DEG_PRECISION)),
+                        phi=FloatEq(math.degrees(math.pi / 2)),
```
rturrado, line 68: Two questions here:
1) Why don't we round `phi`?
2) We wouldn't need to round `theta`, right? Because we know already that `1.23` uses 2 decimals and we allow 5. But, anyway, I would leave it as it makes the code more consistent.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -29,19 +32,32 @@ def __init__(self, qubit_register_name: str):
...
+                acq_protocol="ThresholdedAcquisition",
+            )
+        )
+        return
 
     def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
+        # Note that when adding a rotation gate to the quantify-scheduler Schedule, that there exists an ambiguity with
+        # how quantify-scheduler will store an angle of 180 degrees. Depending on the system the angle may be stored as
+        # either 180 or -180 degrees. This ambiguity has no physical consequences, but may cause the test of the
+        # quantify-scheduler exporter to fail.
```
rturrado, line 50:
```suggestion
        # Note that when adding a rotation gate to the Quantify-scheduler Schedule,
        # there exists an ambiguity with how Quantify-scheduler will store an angle of 180 degrees.
        # Depending on the system the angle may be stored as either 180 or -180 degrees.
        # This ambiguity has no physical consequences, but may cause the exporter test fail.
```

I removed a `that`, because you already start with a `Notice that`.

test/exporter/test_quantify_scheduler_exporter.py
elenbaasc: I'm rounding after converting to degrees, and for the test the pi/2 is converted nicely to 90 degrees, so it does not need to be rounded.

rturrado (approved):

PR #233, opened 2024-06-21
Manage bit (register) variables

rturrado:
Add a bit register to the register manager.
Add a bit side to the measure instructions.
Add a bit register size to the circuit builder constructo.
Change `writer` to write out cQASM 3.0-beta1 compliant code.
Move writers out of `exporter` folder and onto `writer` folder.


opensquirrel/register_manager.py
```diff
@@ -1,24 +1,142 @@
...
+                current_index += 1
+        return cls(register_size, variable_name_to_range, index_to_variable_name)
+
+
+# TODO:
+# In the future, when variables of different types can be defined (e.g. float q)
+# we will have to prevent a variable being called 'q' or 'b'.
+# A possible way to do this would be to introduce variable name mangling in the IR
+# (e.g., store float 'q' as '_float__q')
+# This variable name mangling could be done as an early compiler pass
```
elenbaasc, line 110: What happens now when I define a qubit register 'b' and a bit register 'q'?

opensquirrel/register_manager.py
```diff
@@ -1,24 +1,142 @@
...
+        return self.register_size
+
+    @staticmethod
+    def _parse_ast_string(string: str) -> str:
+        # FIXME: libqasm should return bytes, not the __repr__ of a bytes object ("b'q'")
+        return string[2:-1]
+
+    @classmethod
+    def from_ast(cls, ast: cqasm.semantic.Program) -> Self:
+        register_size: int = 0
```
elenbaasc, line 88: Is it necessary to initialize the (local) variable `register_size` here? It is overwritten in line 93.

opensquirrel/default_measurements.py
```diff
@@ -1,14 +1,14 @@
-from opensquirrel.ir import Measure, Qubit, named_measurement
+from opensquirrel.ir import Bit, Measure, Qubit, named_measurement
 
 
 @named_measurement
-def measure(q: Qubit) -> Measure:
-    return Measure(qubit=q, axis=(0, 0, 1))
+def measure(b: Bit, q: Qubit) -> Measure:
+    return Measure(bit=b, qubit=q, axis=(0, 0, 1))
```
elenbaasc, line 6: I think we should follow the Qiskit convention of the order of the parameters, _i.e._, first the qubit that is measure and then the bit into which the measurement outcome is stored: see https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#measure

opensquirrel/ir.py
```diff
@@ -189,21 +206,22 @@ class Statement(IRNode, ABC):
 
 
 class Measure(Statement, ABC):
-
     def __init__(
         self,
+        bit: Bit,
```
elenbaasc, line 211: I think we should follow the Qiskit convention of the order of the parameters, _i.e._, first the qubit that is measure and then the bit into which the measurement outcome is stored: see https://docs.quantum.ibm.com/api/qiskit/qiskit.circuit.QuantumCircuit#measure

opensquirrel/ir.py
```diff
@@ -388,6 +406,7 @@ def named_gate(gate_generator: Callable[..., MatrixGate]) -> Callable[..., Matri
 def named_gate(gate_generator: Callable[..., ControlledGate]) -> Callable[..., ControlledGate]: ...
 
 
+# TODO: can we extract the duplicated code for the two wrapper functions below out to a single function?
```
elenbaasc, line 409: Please create an issue for this. We should limit the amount of `TODO`s in the code.

elenbaasc:
Nice work! Hardly any comments 😃

opensquirrel/reindexer/qubit_reindexer.py
```diff
@@ -3,8 +3,8 @@
 from collections.abc import Iterable
 from typing import TYPE_CHECKING
 
-from opensquirrel.ir import IR, BlochSphereRotation, ControlledGate, Gate, IRVisitor, MatrixGate, Measure, Qubit
-from opensquirrel.register_manager import RegisterManager
+from opensquirrel.ir import IR, Bit, BlochSphereRotation, ControlledGate, Gate, IRVisitor, MatrixGate, Measure, Qubit
```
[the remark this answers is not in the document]
rturrado: Good spot, Chris, thanks!

opensquirrel/register_manager.py
rturrado: So this would be the scenario now:
- You create a correct program with a qubit `b` and a bit `q`.
- OpenSquirrel parses the semantic AST coming from libqasm and creates a qubit register `q` and a bit register `b` with the needed amount of qubits and bits respectively.
- The qubit `b` and the bit `q` variables are no longer used. All the code in OpenSquirrel works directly with the qubit register `q` and the bit register `b` through the RegisterManager. For example, a gate such as `X b`, operating on qubit variable `b` will be internally processed as an `X` gate with a `Qubit` operand with a  given index, and that index is a position in the qubit register `q`.
- The RegisterManager still keeps data structures though connecting variable names to ranges in the qubit/bit registers. For example, it could tell you that qubit  variable `b` sits in index 5 of the qubit register `q`.

rturrado: Very good question though!

opensquirrel/default_measurements.py
rturrado: I agree. I did it this way because the semantic AST has a bit first operand and a qubit second operand. But that's not strong enough a reason to keep that order in OpenSquirrel. I'll put bits behind. Thanks!

opensquirrel/ir.py
rturrado: Good idea, thanks!

rturrado: Issue created: https://github.com/QuTech-Delft/OpenSquirrel/issues/234

opensquirrel/register_manager.py
rturrado: Done!

opensquirrel/default_measurements.py
rturrado: Done!

opensquirrel/ir.py
rturrado: Done!

opensquirrel/register_manager.py
elenbaasc: Good to know that it doesn't break. Should we add a unit test with this scenario?

rturrado: Good idea, I'll add one!

rturrado: I have added `test_qubit_variable_b_and_bit_variable_q`.

elenbaasc (approved):

PR #235, opened 2024-06-24
Add cQASM Lite writer

rturrado:
Add `cqasm_lite_writer` to write out "cQASM Lite" code (cQASM 3.-beta1 code without 'bits' references).


elenbaasc (approved):
Apart from the fact that we might want to add tests to the `test_writer`  with input circuits that have bits and measurements and that these are 'preserved' as expected, I think it should be fine. (Also, because these tests are already part of other tests as well, they might not be strictly necessary for now.)

So, nice work! Great that we could finish this today 😄

rturrado:
Thanks!

Yes, I didn't add them because we didn't have them previously, let's say, and as you point, the writer is well exercised, for example, in the integration tests.

PR #250, opened 2024-07-03
Make opensquirrel compatible with numpy 2.0

eendebakpt:
Replace numpy deprecated (and in numpy 2.0 removed) functionality with modern numpy options

### DIFF
--- a/opensquirrel/common.py
+++ b/opensquirrel/common.py
@@ -17,7 +17,7 @@ def normalize_angle(x: float) -> float:
     return t
 
 
-def are_matrices_equivalent_up_to_global_phase(matrix_a: NDArray[np.complex_], matrix_b: NDArray[np.complex_]) -> bool:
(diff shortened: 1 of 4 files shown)


juanboschero (dismissed):

elenbaasc:
We have a ticket to check what is required (and what the consequences will be) to update to numpy 2.0. We will pick up these changes when we tackle that.

eendebakpt:
> We have a ticket to check what is required (and what the consequences will be) to update to numpy 2.0. We will pick up these changes when we tackle that.

Nice! Note even without numpy 2.0 support these changes can be worthwhile as they replace methods that are also deprecated in nump 1.26.x

rturrado (approved):
Beautiful! Thanks for the changes, Peter.

S-Linde (dismissed):

PR #253, opened 2024-07-08
[CQT-30] Print out available information instead of "anonymous-gate"

elenbaasc:


opensquirrel/ir.py
```diff
@@ -355,7 +363,7 @@ def __init__(
         self.operands = operands
 
     def __repr__(self) -> str:
-        return f"MatrixGate(qubits={self.operands}, matrix={self.matrix})"
+        return f"MatrixGate(qubits={self.operands}, matrix={np.round(self.matrix, self._significant_digits_repr)})"
```
juanboschero, line 366: Is it wise to represent the matrix of a Matrixgate? If there are many operands involved, this may bloat the log.

test/writer/test_writer.py
```diff
@@ -36,21 +39,26 @@ def test_write() -> None:
...
 
 qubit[2] q
+bit[2] b
 
-CR(1.234) q[0], q[1]
-<anonymous-gate>
+H q[0]
+BlochSphereRotation(Qubit[0], axis=[0.57735 0.57735 0.57735], angle=1.23, phase=0.0)
+ControlledGate(control_qubit=Qubit[0], BlochSphereRotation(Qubit[0], axis=[0.57735 0.57735 0.57735], angle=1.23, phase=0.0))
+MatrixGate(qubits=[Qubit[0], Qubit[1]], matrix=[[1 0 0 0]
```
juanboschero, line 58: In order to abide by the CQASM standards, we may want to have our commands inline, even if it won't be parsed through CQASM.

elenbaasc: Agreed! I've made sure that it outputs to a single line.

opensquirrel/ir.py
elenbaasc: This could indeed be a large output, but it is mostly for diagnostic purposes, so I would like to provide all the available information to the user.

opensquirrel/ir.py
```diff
@@ -309,7 +314,10 @@ def identity(q: Qubit) -> BlochSphereRotation:
         return BlochSphereRotation(qubit=q, axis=(1, 0, 0), angle=0, phase=0)
 
     def __repr__(self) -> str:
-        return f"BlochSphereRotation({self.qubit}, axis={self.axis}, angle={self.angle}, phase={self.phase})"
+        axis = np.round(self.axis, self._significant_digits_repr)
+        angle = np.round(self.angle, self._significant_digits_repr)
+        phase = np.round(self.phase, self._significant_digits_repr)
```
rturrado, line 319: I'd recommend to have a `_round(f: float): float` function that would just `return np.round(f, self._significant_digits_repr)`, then rewrite all the calls to `np.round(blah, self._significant_digits_repr)` as `self._round(blah)`.

opensquirrel/writer/writer.py
```diff
@@ -42,6 +42,7 @@ def visit_measure(self, measure: Measure) -> None:
     def visit_gate(self, gate: Gate) -> None:
         gate_name = gate.name
         if gate.is_anonymous:
+            gate_name = gate_name.replace("\n", "")
             self.output += f"{gate_name}\n"
```
rturrado, line 46: This code looks  quite weird. If you read it, you don't understand why all newlines are removed.

When implementing a string representation for a node, I'd recommend never to add a trailing newline in the callee, but in the caller. So, ultimately, if you want to print the whole IR, you would do something like `print(f"{}\n", ir). This way, you avoid the cases when you would like to get rid of the trailing newline.

opensquirrel/ir.py
elenbaasc: Could point! Implemented 😄

opensquirrel/writer/writer.py
elenbaasc: In the case of  a MatrixGate we want the newline escapes (\n) to be removed from the string because otherwise the array of the matrix does not print on a single line. I have added this explanation to the code.

opensquirrel/ir.py
```diff
@@ -247,6 +247,8 @@ def get_qubit_operands(self) -> list[Qubit]:
 
 
 class Gate(Statement, ABC):
+    _significant_digits_repr = 5
```
rturrado, line 250: Ideally, I would have the `_significant_digits_repr` variable just at `__repr__` scope, and not as a Gate member. However, I see it's used in the subclasses. I would:
1. Change `_round` to receive a `decimals=self.repr_decimals`, just as `np.round` [does](https://numpy.org/doc/stable/reference/generated/numpy.round.html), and then call `np.round(value, decimals)`. The only reason to keep `_round` and not to use `np.round` directly is that `_round` manages other types.
2. Change `_significant_digits_repr` to `repr_decimals` (as in, repr.decimals, or decimals used in `__repr__`). Just to be consistent with `np.round` naming. I think it's simpler and clearer.

opensquirrel/writer/writer.py
```diff
@@ -42,6 +42,10 @@ def visit_measure(self, measure: Measure) -> None:
     def visit_gate(self, gate: Gate) -> None:
         gate_name = gate.name
         if gate.is_anonymous:
+            if "MatrixGate" in gate_name:
+                # In the case of a MatrixGate the newline escapes (\n) should be removed from the array such that the
+                # array is printed on a single line.
```
rturrado, line 47:
```suggestion
                # In the case of a MatrixGate the newlines should be removed from the array
                # such that the array is printed on a single line.
```

test/writer/test_writer.py
```diff
@@ -36,21 +39,23 @@ def test_write() -> None:
...
         == """version 3.0
 
 qubit[2] q
+bit[2] b
 
-CR(1.234) q[0], q[1]
-<anonymous-gate>
+H q[0]
+Anonymous gate: BlochSphereRotation(Qubit[0], axis=[0.57735 0.57735 0.57735], angle=1.23, phase=0.0)
+Anonymous gate: ControlledGate(control_qubit=Qubit[0], BlochSphereRotation(Qubit[0], axis=[0.57735 0.57735 0.57735], angle=1.23, phase=0.0))
```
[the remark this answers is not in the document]
elenbaasc: As you stated in your message on Slack, indeed a controlled gate is printed as an anonymous gate if the target get is anonymous or the entire controlled gate is anonymous, e.g. the CY gate is anonymous, even though the Y gate is not.

opensquirrel/ir.py
elenbaasc: I've moved the helperfunction `_repr_round` and variable `_REPR_DECIMALS_` (used by the BSR and MatrixGate) to the global scope.

opensquirrel/ir.py
```diff
@@ -13,6 +13,15 @@
...
 
+_REPR_DECIMALS = 5
+
+
+def _repr_round(
+        value: float | Axis | NDArray[np.complex64],
+        decimals: int = _REPR_DECIMALS) -> float | Axis | NDArray[np.complex64]:
+    return np.round(value, decimals)
+
+
```
rturrado, line 24: I leave this up to you, but I would move this code to utils/math.py, and I would add a few basic tests for this function: round a float, an axis, and an NDArray, and pass no decimals, 0, 1, and -1.

rturrado: If you check the `np.round` docs, this is what they say for a negative decimal input:

> If decimals is negative, it specifies the number of positions to the left of the decimal point.

opensquirrel/ir.py
rturrado: Can you please remove the leading underscore? I would use it as a hint to the reader that a class member/method is intended to be private, but not here.

opensquirrel/ir.py
rturrado, line 24: Can you please remove the leading underscores? I would use them as a hint to the reader that a class member/method is intended to be private, but not here.

rturrado (approved):
Ready to go!

Great work!

PR #256, opened 2024-07-15
Dont raise bare exceptions

### LINKED ISSUE #125: Don't raise bare excpetions
Right now there are quite a few places in opensquirel where exceptions of class `Exception` are raised. This makes it impossible for consumers of opensquirrel to catch these exceptions as being Opensquirel specific.

Could we please make sure that all raised exceptions by opensquirel are from a subclass so that it is easy to handle opensquirel related exceptions, and that we don't need the very generic `except Exception` in our downstream code which catches all exceptions, which is quite bad.

S-Linde:
This merge request fixes incorrect error handling. No bare exceptions are raised anywhere in the code. 
Furthermore, all `assert` statements are removed except those in the test folders and the `check_passes` subpackage.

Closes #125


opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -61,14 +60,23 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
...
+            f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {g}. "
+            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
+        )
 
     def visit_matrix_gate(self, g: MatrixGate) -> None:
-        raise _unsupported_gates_exception
+        raise UnsupportedGatesError(
+            f"Cannot export circuit: it contains unsupported gates. MatrixGates are not supported. "
+            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
+        )
```
rturrado, line 72: Can we make a constructor for UnsupportedGatesError receiving a string?

The constructor would be something like:
```
    def __init__(self, gate_type: str):
        self.message =  f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {gate_type}"
            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
```

Then, from `visit_matrix_gate`, we would call `raise UnsupportedGatesError('MatrixGate')`.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -61,14 +60,23 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
...
 
     def visit_controlled_gate(self, g: ControlledGate) -> None:
         if not isinstance(g.target_gate, BlochSphereRotation):
-            raise _unsupported_gates_exception
+            raise UnsupportedGatesError(
+                f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {g}. "
+                "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
+            )
```
rturrado, line 79: And from here we would call with `raise UnsupportedGatesError(f"{g}")`, or `str(g)`...

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -86,15 +94,18 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
             )
             return
 
-        raise _unsupported_gates_exception
+        raise UnsupportedGatesError(
+            f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {g}. "
+            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
+        )
```
rturrado, line 100: And the same here.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -6,18 +6,18 @@
...
     pass
 
-_unsupported_gates_exception = Exception(
-    "Cannot export circuit: it contains unsupported gates. "
-    "Decompose them to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
-)
+
+class UnsupportedGatesError(Exception):
+    """Should be raised when the set of gates is not supported by the exporter."""
+
```
rturrado, line 20:
```suggestion

class UnsupportedGateError(Exception):
    """Should be raised when the gate is not supported by the exporter."""

```

Because every time we raise it we will be complaining about a specific gate, right?

opensquirrel/utils/matrix_expander.py
```diff
@@ -95,19 +95,22 @@ def __init__(self, qubit_register_size: int) -> None:
         self.qubit_register_size = qubit_register_size
 
     def visit_bloch_sphere_rotation(self, rot: BlochSphereRotation) -> NDArray[np.complex_]:
-        assert rot.qubit.index < self.qubit_register_size
+        if rot.qubit.index >= self.qubit_register_size:
+            raise IndexError("Index out of range.")
```
rturrado, line 99: I'm more than fine with these changes. I know Pablo differentiated between using assert and exceptions. In C++ code, asserts are a debug mechanism. They are kind of extra checks. If you have a runtime error, and you debug it, these extra checks give you more information. But I am not really used to program with asserts, and I prefer these let's say hard checks that are done also in release mode, via exceptions.

rturrado (changes requested):
Great work!

All good from my side except for a couple of minor things:
1. I would change the name of the exception to singular. The plural would make sense if you collect many errors and then you raise them all at once.
2. I would add an exception constructor so that we don't repeat most of the error message time and again. We can just pass the unsupported gate name.

opensquirrel/exporter/quantify_scheduler_exporter.py
S-Linde: In this case the error message is specifically tailored towards Quantify Scheduler, but the error type `UnsupportedGatesError ` is more general. I like the idea of making a more standardized and easier way to use the `Exception`, but I think the place should be in the `_ScheduleCreator` (which we might want to rename to `_QuantifyScheduleCreator`).

So, something like
```python
class _ScheduleCreator:
  # ----------------
  # Other methods
  # ----------------

  @staticmethod
  def _raise_unsupported_gates_error(gate_type: Any) -> None:
    raise UnsupportedGatesError(
            f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {gate_type}"
            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
        )
```

rturrado: Hm... I think we should decouple Error classes from Scheduler classes.

I see OK to have a generic UnsupportedGateError, and maybe then a QSUnsupportedGateError.

The generic class could always raise `f"Cannot export circuit: it contains an unsupported gate: {message}"`, then QSUnsupportedGateError could raise `"{gate_type}. Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)`.

But the code would just do `raise QSUnsupportedGateError(gate_type)`.

opensquirrel/exporter/quantify_scheduler_exporter.py
S-Linde: You're right, its fixed now.

opensquirrel/exporter/quantify_scheduler_exporter.py
S-Linde: I made a new implementation were `UnsupportedGateError` is now easily initialized as you suggested. If it is raised by the exporter, than it will now raise an `ExportError` from the `UnsupportedGateError` with a message tailored towards the specific exporter (we currently have only 1)

opensquirrel/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+class UnsupportedGateError(Exception):
+    """Should be raised when a gate is not supported."""
+
+    def __init__(self, gate: Any, *args: Any) -> None:
+        """Init of the ``UnsupportedGateError``.
+
+        Args:
+            gate: Gate that is not supported.
+        """
+        self.unsupported_gate = gate
```
rturrado, line 15: I wouldn't keep `unsupported_gate` as a member, but just the message that is passed  to `super().__init__`.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -86,21 +82,30 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
             )
             return
 
-        raise _unsupported_gates_exception
+        raise UnsupportedGateError(
+            f"Cannot export circuit: it contains unsupported gates. The following gate could not be exported: {g}. "
+            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
+        )
```
rturrado, line 88:
```suggestion
        raise UnsupportedGateError(g)
```

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -86,21 +82,30 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
...
         quantify_scheduler_gates = QuantifySchedulerNotInstalled()
 
     schedule_creator = _ScheduleCreator(circuit.qubit_register_name)
-    circuit.ir.accept(schedule_creator)
+    try:
+        circuit.ir.accept(schedule_creator)
+    except UnsupportedGateError as e:
+        raise ExportError(
+            "Cannot export circuit: it contains unsupported gates. The following gate could not be exported: "
+            f"{e.unsupported_gate}. Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
```
rturrado, line 109:
```suggestion
            f"cannot export circuit: {e}. Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
```

I would just print the error message:
- The `it contains unsupported gates` text is redundant.
- The `The following gate could not be exported: ` should como from `UnsupportedGateError` class.
- All in all, we could read: `cannot export circuit: the following gate is not supported: blah. Decompose...`
- This way, we could also reuse the same message for different types of exceptions. In case that couldn't be possible (e.g., the `Decompose all gates...` part could only make sense for UnsupportedGateError exceptions), then we could create a QuantifySchedulerUnsupportedGateError that would add that last piece of advise.

Note also that error messages should start in lowercase, because they will usually be suffixed (with `Error: `, or `Cannot export circuit: `, etc.)

opensquirrel/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+        super().__init__(f"The following gate is not supported: {gate}.", *args)
+
```
rturrado, line 17:
```suggestion
        super().__init__(f"the following gate is not supported: {gate}.", *args)

```

Error messages should start in lowercase, because they will usually be suffixed (with `Error: `, or `Cannot export circuit: `, etc.)

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -86,21 +82,30 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
...
         quantify_scheduler = QuantifySchedulerNotInstalled()
         global quantify_scheduler_gates
         quantify_scheduler_gates = QuantifySchedulerNotInstalled()
 
     schedule_creator = _ScheduleCreator(circuit.qubit_register_name)
-    circuit.ir.accept(schedule_creator)
+    try:
+        circuit.ir.accept(schedule_creator)
+    except UnsupportedGateError as e:
+        raise ExportError(
```
rturrado, line 107: Not sure about the `ExportError` name:
- I would rather name it `QuantifySchedulerExporterError`.
- `ImportError` is a Python specific error regarding import naming.
- `ExportError` can be confusing in the sense that one would associate it with `ImportError`, when it actually has nothing to do with it, and may even think that it is a Python specific error as well.
- `QuantifySchedulerExporterError` leaves no space to doubt.

rturrado (changes requested):
Hi Stan,

I've added some final comments:
- Keep Error clases to store only error messages.
- Just reuse those error messages, without duplicating any other text (e.g. when creating an ExportError).
- Rename ExportError to QuantifySchedulerExporterError.
- Error messages should start in lowercase, and not end in any kind of punctuation. That makes them suitable for inserting in other error messages. This is probably more debatable, but it is the way I'm used to work in C++, and the standard for Rust as well, although may not be the case in Python.

Anyway, think of the ExportError case: you put the UnsupportedGateError in the middle of another message (which is something very common). So you can end up with:
- `Cannot export circuit: The gate is not supported...`, or
- `Cannot export circuit: the gate is not supported...`.

And that should apply to whoever catches an ExportError too:
- `Error: Cannot export...`, or
- `Error: cannot export...`.

opensquirrel/exporter/quantify_scheduler_exporter.py
rturrado: Thanks!

opensquirrel/exporter/quantify_scheduler_exporter.py
rturrado: I think you just missed this one.

opensquirrel/exporter/quantify_scheduler_exporter.py
S-Linde: I would prefer to have the name `QuantifyScheduler` out of the name of the Exception class. What do you think of `CircuitExportError`?

S-Linde:
> Hi Stan,
> 
> I've added some final comments:
> 
> * Keep Error clases to store only error messages.
> * Just reuse those error messages, without duplicating any other text (e.g. when creating an ExportError).
> * Rename ExportError to QuantifySchedulerExporterError.
> * Error messages should start in lowercase, and not end in any kind of punctuation. That makes them suitable for inserting in other error messages. This is probably more debatable, but it is the way I'm used to work in C++, and the standard for Rust as well, although may not be the case in Python.
> 
> Anyway, think of the ExportError case: you put the UnsupportedGateError in the middle of another message (which is something very common). So you can end up with:
> 
> * `Cannot export circuit: The gate is not supported...`, or
> * `Cannot export circuit: the gate is not supported...`.
> 
> And that should apply to whoever catches an ExportError too:
> 
> * `Error: Cannot export...`, or
> * `Error: cannot export...`.

Thank you for the detailed explanation. The error message of `UnsuportedGateError` is now reused and it starts with a lower case and has no punctuation at the end.

opensquirrel/exporter/quantify_scheduler_exporter.py
S-Linde: I changed the error message to be a bit more concise. However, I do think that `Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"` is relevant here, because we only raise the error with this specific message when we catch an `UnsupportedGateError`. I would really prefer to keep the Error types to general concepts, and the messages really precise.

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -86,21 +82,27 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
...
         quantify_scheduler_gates = QuantifySchedulerNotInstalled()
 
     schedule_creator = _ScheduleCreator(circuit.qubit_register_name)
-    circuit.ir.accept(schedule_creator)
+    try:
+        circuit.ir.accept(schedule_creator)
+    except UnsupportedGateError as e:
+        raise CircuitExportError(
+            f"cannot export circuit: {e} Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, "
+            "cz)"
```
rturrado, line 106:
```suggestion
            f"cannot export circuit: {e}. "
            "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
```

Split into two self-contained lines just for reader's clarity.
And added a dot after the error message.

opensquirrel/exporter/quantify_scheduler_exporter.py
rturrado: Yes, yes, I wouldn't remove the `Decompose all gates...` part. Thanks!

opensquirrel/exporter/quantify_scheduler_exporter.py
rturrado: I suggested to include `QuantifySchedulerExporter` in the name because this code belongs to Quantify scheduler. But I agree with having a more generic name. I would then use `ExporterError`.

All this naming becomes clearer when you put all error classes together in a file. Then you polish the names looking for consistency.

rturrado (commented):
A couple of questions, Stan:

1. Are you going to put all the error classes in a file as part of this PR? Maybe worth doing as part of another PR.
2. Also maybe worth doing as part of another PR (if you agree that we should follow the convention of error messages starting in lowercase and not ending in a punctuation sign), would be to revise all the error messages in the code to follow this convention. I wonder if that could be enforced via a linter as well.

Thanks! For this last review I've just done two minimal comments.

opensquirrel/merger/general_merger.py
```diff
@@ -16,7 +16,7 @@ def compose_bloch_sphere_rotations(a: BlochSphereRotation, b: BlochSphereRotatio
     Uses Rodrigues' rotation formula, see for instance https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula.
     """
     if a.qubit != b.qubit:
-        raise ValueError("Cannot merger two BlochSphereRotation's on different qubits")
+        raise ValueError("cannot merger two BlochSphereRotation's on different qubits")
```
rturrado, line 19:
```suggestion
        raise ValueError("cannot merge two BlochSphereRotation's on different qubits")
```

rturrado (changes requested):
Great work Stan!

There is only a minimal fix for a typo left.

Thanks!

rturrado (dismissed):

elenbaasc (approved):

PR #265, opened 2024-07-25
Cqt 76 linked ir between circuit and circuit builder

S-Linde:
Use `deepcopy` to copy the `IR` and `RegisterManager` of the builder instead of using a reference.
Added a test for this behavior as well.

### DIFF
--- a/test/test_circuit_builder.py
+++ b/test/test_circuit_builder.py
@@ -121,5 +121,5 @@ class TestCircuitBuilder:
     def test_decoupling_circuit_and_builder(self) -> None:
         builder = CircuitBuilder(1)
         circuit = builder.to_circuit()
-        builder.I(Qubit(0))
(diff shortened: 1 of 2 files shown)


test/test_circuit_builder.py
juanboschero, line 1: Nice fix Stan! I would add a further assertion to the test:
```
assert isinstance(circuit.__repr__(), str)
```

juanboschero (changes requested):

test/test_circuit_builder.py
S-Linde: Wouldn't this already be caught by `mypy`?

juanboschero: It was also to test the writer with the deepcopy but I think you are right, I think mypy may flag it before hand.

PR #269, opened 2024-07-26
CQT 39: check if ANTLR4 dependency is still required (otherwise, remove)

rturrado:
Remove ANTLR4 dependency.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,7 +30,6 @@ packages = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
-antlr4-python3-runtime = "^4.13.1"
 numpy = [
     { version = "1.24", python = "<3.9" },
     { version = "^1.26", python = "^3.9" },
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -80,7 +79,7 @@ asyncio_mode = "auto"
 
 [tool.coverage.run]
 branch = true
-omit = ["tests/*", "**/.tox/**", "opensquirrel/parsing/antlr/generated/*"]
```
elenbaasc, line 83: Good catch 😄

elenbaasc (commented):
Do you know if we need to update the `poetry.lock` file and commit the changes, before merging?

rturrado:
> Do you know if we need to update the `poetry.lock` file and commit the changes, before merging?

I don't think `poetry.lock` should be part of the codebase, should it?

rturrado:
> > Do you know if we need to update the `poetry.lock` file and commit the changes, before merging?
> 
> I don't think `poetry.lock` should be part of the codebase, should it?

It seems it should: https://python-poetry.org/docs/basic-usage/#committing-your-poetrylock-file-to-version-control

I've removed the ANTLR4 package from `poetry.lock` as well.

elenbaasc (approved):
Cool! Can be merged :)

PR #273, opened 2024-08-01
[CQT-58] Implement and configure tox

S-Linde:
This pull request configures `tox`.

Testing and and type checking remain the same, but linting has been altered. Linting now uses `ruff`, with an extended list of selected linters options. All options that we currently do not comply with are commented out, but should be turned on at some point. This is an extended list indeed, but if we manage to comply with these rules we will have a uniform and well thought out codebase. More importantly, most style related issues will be resolved before a pull request is made, meaning that a review can focus much more on content than style.

Note that we don't use `isort`, since `ruff` can sort the imports for us.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,9 +158,11 @@ flake8-type-checking.exempt-modules = ["typing", "typing_extensions", "numpy", "
     "S101", # asserts allowed in tests
     "FBT", # Don't care about booleans as positional arguments in tests, e.g. via @pytest.mark.parametrize()
     "D", # No doc requirements in tests
+    "SLF001", # Testing private functions is ok
 ]
 "check_*.py" = [
     "S101", # asserts allowed in check functions
+    "SLF001", # Checking private functions is ok
 ]
 
 [tool.black]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,7 @@ libqasm = "0.6.6"
 networkx = "^3.0.0"
 
 [tool.poetry.group.dev.dependencies]
+black = ">=23.11,<25.0"
 coverage = {extras = ["toml"], version = "^7.3.2"}
 pytest = {extras = ["toml"], version = ">=7.4.3,<9.0.0"}
 pytest-cov = ">=4.1,<6.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -98,24 +98,32 @@ target-version = "py38"
 [tool.ruff.lint]
 select = [
     "A", # flake8-builtins
-    #"B", #flake8-bugbear
+    #"ANN", # flake8-annotations
+    "ASYNC", # flake8-async
+    #"B", # flake8-bugbear
+    "BLE", # flake8-blind-except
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -91,12 +91,72 @@ exclude_lines = [
...
+]
+isort.combine-as-imports = true
+pydocstyle.convention = "google"
+flake8-type-checking.exempt-modules = ["typing", "typing_extensions", "numpy", "networkx", "collections"]
+
+[tool.ruff.lint.extend-per-file-ignores]
+"test/**/*.py" = [
+    "S101", # asserts allowed in tests
+    "FBT", # Don't care about booleans as positional arguments in tests, e.g. via @pytest.mark.parametrize()
+    "D", # No doc requirements in tests
```
juanboschero, line 155: `"W605" # Dont worry about invalid escape sequences (deprecated in Python 3.6)`

juanboschero (commented):
Great work, excited for ruff!

tox.ini
```diff
@@ -28,10 +27,10 @@ commands =
 description = run fixing linters
 commands =
     poetry run black .
-    poetry run isort .
+    poetry run ruff check --fix
```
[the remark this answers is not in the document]
S-Linde: This is something we can do in the future, when we turn on more of the linting of `ruff`. Currently we need black to check the formatting, so we don't lose that check.

pyproject.toml
S-Linde: Nice catch, added it to the ignore list.

juanboschero (dismissed):

tox.ini
```diff
@@ -27,11 +25,10 @@ commands =
...
-    poetry run isort .
-    ; poetry run docformatter -i --black --recursive --wrap-summaries=119 --wrap-descriptions=119 .
+    poetry run ruff check --fix
+    poetry run ruff format
 
 [testenv:test]
 description = run unit tests
 commands =
-    poetry run pytest .
+    poetry run pytest . -vv
```
rturrado, line 34: Nice!

rturrado (dismissed):

rturrado:
@juanboschero @S-Linde The current lint error triggers an interesting error! `_find_unused_index` is a private method, and we are accessing it through a test. This simply wouldn't compile in C++, but it is allowed in Python (linters aside). The solution to this, instead of relaxing the linter, may well be to get rid of the tests, and let the private method be tested via the public method that uses it.

S-Linde:
> @S-Linde @juanboschero The current lint error triggers an interesting question! `_find_unused_index` is a private method, and we are accessing it through a test. This simply wouldn't compile in C++, but it is allowed in Python (linters aside). The solution to this, instead of relaxing the linter, may well be to get rid of the tests, and let the private method be tested via the public method that uses it.

I personally prefer to be able to check private methods. Private methods often contain the actual implementation of an algorithm or a small, but more easily tested part of an algorithm. I don't think all private methods should be tested, but I do think it should be allowed.

juanboschero (approved):

PR #279, opened 2024-08-06
Remove superfluous rounding in decomposition

juanboschero:


opensquirrel/common.py
```diff
@@ -18,11 +18,16 @@ def normalize_angle(x: float) -> float:
 
 
 def are_matrices_equivalent_up_to_global_phase(matrix_a: NDArray[np.complex_], matrix_b: NDArray[np.complex_]) -> bool:
-    order_magnitude = abs(math.floor(math.log10(ATOL)))
+    """
+    Checks whether two matrices are equivalent up to global phase.
```
elenbaasc, line 22:
```suggestion
    Checks whether two matrices are equivalent up to a global phase.
```

opensquirrel/common.py
```diff
@@ -18,11 +18,16 @@ def normalize_angle(x: float) -> float:
...
 
-    matrix_a = np.round(matrix_a, order_magnitude)
-    matrix_b = np.round(matrix_b, order_magnitude)
+    Args:
+        matrix_a: first matrix.
+        matrix_b: second matrix.
 
+    Returns:
+        Whether two matrices are equivalent up to both the ATOL and the global phase.
```
elenbaasc, line 29: ATOL is simply our defined tolerance, we assume that anything that is equal within ATOL _is_ equal.

```suggestion
        Whether two matrices are equivalent up to a global phase.
```

elenbaasc (commented):
Thanks for checking! Only very minor comments.

elenbaasc (approved):

PR #281, opened 2024-08-08
[CQT-50] Add extra tests to writers where needed

rturrado:
Add a few tests.
Rename also some tests.
Rename `num_significant_digits` to `FLOAT_PRECISION` (for consistency with Quantify Scheduler exporter's `DEG_PRECISION`).
Change writer to always only output one single trailing newline.

### DIFF
--- a/opensquirrel/writer/writer.py
+++ b/opensquirrel/writer/writer.py
@@ -4,6 +4,7 @@ from opensquirrel.register_manager import RegisterManager
 
 
 class _WriterImpl(IRVisitor):
+    # Precision used when writing out a float number
     FLOAT_PRECISION = 8
 
     def __init__(self, register_manager: RegisterManager) -> None:
--- a/opensquirrel/writer/writer.py
+++ b/opensquirrel/writer/writer.py
@@ -64,4 +64,4 @@ def circuit_to_string(circuit: Circuit) -> str:
 
     circuit.ir.accept(writer_impl)
 
-    return writer_impl.output.rstrip() + "\n"  # remove all trailing lines, and leave only one
+    return writer_impl.output.rstrip() + "\n"  # remove all trailing lines and leave only one
--- a/opensquirrel/writer/writer.py
+++ b/opensquirrel/writer/writer.py
@@ -63,8 +63,4 @@ def circuit_to_string(circuit: Circuit) -> str:
 
     circuit.ir.accept(writer_impl)
 
-    return (
-        writer_impl.output
-        .rstrip()  # remove all trailing lines
-        + '\n'  # and leave only one
-    )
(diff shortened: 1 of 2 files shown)


opensquirrel/writer/writer.py
```diff
@@ -63,4 +63,4 @@ def circuit_to_string(circuit: Circuit) -> str:
 
     circuit.ir.accept(writer_impl)
 
-    return writer_impl.output
+    return writer_impl.output.rstrip() + "\n"  # remove all trailing lines, and leave only one
```
juanboschero, line 66:
```suggestion
    return writer_impl.output.rstrip() + "\n"  # remove all trailing lines and leave only one
```

test/writer/test_writer.py
```diff
@@ -1,27 +1,53 @@
...
+    assert (
+        writer.circuit_to_string(circuit)
+        == """version 3.0
 
+qubit[3] q
+"""
+    )
+
+
+def test_circuit_with_qubits_and_bits() -> None:
```
juanboschero, line 20: I don't see where the bits are being tested in this method. Maybe a better name would be `test_circuit_with_control_operations()`

juanboschero (changes requested):

opensquirrel/writer/writer.py
```diff
@@ -4,7 +4,7 @@
 
 
 class _WriterImpl(IRVisitor):
-    number_of_significant_digits = 8
+    FLOAT_PRECISION = 8
```
S-Linde, line 7: Shouldn't his be `FIXED_POINT_PRECISION`, since we are using this to convert the floating point to a fixed point?

test/writer/test_writer.py
rturrado: Very well spotted! Fixed now.

opensquirrel/writer/writer.py
rturrado: Nice, thanks!

opensquirrel/writer/writer.py
rturrado: I don't think so.

1. We are not converting anything. We are just formatting an output with a given precision.
2. Fixed point and floating point are two different formats for representing a real number. But here we are dealing with `IR.Float`, which contains a Python's `float`, which is a floating point number. So any reference to fixed point is misleading here, IMO.

Python's `round` function names this precision argument as `ndigits`. numpy uses the term `decimals`. Quantify Scheduler exporter uses the term `DEG_PRECISION` (radian to degree conversion outcome precision)....

What do you think of calling it `FLOAT_OUTPUT_DECIMALS`, or `FLOAT_OUTPUT_NDIGITS`?

juanboschero (approved):

Issue #210, opened 2024-05-31
Allow list in Measure function


juanboschero:
The current `Measure` function allows a single Qubit object i.e `Measure(Qubit(1))`. If one wants multiple measures, the user needs to specify:
```
Measure(Qubit(0))
Measure(Qubit(1))
Measure(Qubit(2))
```

1. Allow Measure to accept a list `Measure([Qubit(0),Qubit(1),Qubit(2)])`
2. Allow Measure to accept indices `Measure(qubits = [0,1,2])`
3. Allow measure all functionality when user does not specify a qubit, i.e `Measure()` performs a measure all

rturrado:
Regarding point 3, we have to take care because the language doesn't accept that form.

I think we should keep consistency between the language and the CircuitBuilder.

rturrado:
@elenbaasc @juanboschero I think we have to be more specific about this issue. I think the idea would be that, as well as you can use SGMQ (Single Gate Multiple Qubit) notation in the language, you should also be able to use it in the CircuitBuilder. But **that doesn't mean that the Measure IR node can have multiple qubits**. I think that would complicate the work on the rest of the compiler passes (imagine treating a measure q[1, 3, 5] as a single node instead of as 3 separate nodes, in terms of moving instructions around, dependencies with other instructions and so on...).

PR #311, opened 2024-09-02
#210 : Allow list in Measure function

arcturusannamalai:
- support the 1, 2 requests from the issue but not a empty Measure() object.
- updates repr() test for Measure object

### DIFF
--- a/opensquirrel/default_measurements.py
+++ b/opensquirrel/default_measurements.py
@@ -3,12 +3,13 @@ from opensquirrel.ir import Bit, Measure, Qubit, named_measurement
 
 @named_measurement
 def measure(q: Qubit, b: Bit) -> Measure:
-    return Measure(qubit=q, bit=b, axis=(0, 0, 1))
+    return Measure(qubits=[q.index, ], bit=b, axis=(0, 0, 1))
+    #return Measure(qubits=[q,], bit=b, axis=(0, 0, 1))
 
 
 @named_measurement
 def measure_z(q: Qubit, b: Bit) -> Measure:
-    return Measure(qubit=q, bit=b, axis=(0, 0, 1))
+    return Measure(qubits=[q,], bit=b, axis=(0, 0, 1))
 
 
 default_measurement_set = [measure_z, measure]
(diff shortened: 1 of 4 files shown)


rturrado:
Thanks for the PR @arcturusannamalai.

I wouldn't say not meeting point 3 is important at the moment. Actually, it is desirable, as I think we have to keep consistency between the language and the CircuitBuilder.

@elenbaasc @juanboschero I think we have to be more specific about this issue. I think the idea would be that, as well as you can use SGMQ (Single Gate Multiple Qubit) notation in the language, you should also be able to use it in the CircuitBuilder. But that doesn't mean that the Measure IR node can have multiple qubits (**which is what is being done in this PR**). I think that would complicate the work on the rest of the compiler passes (imagine treating a `measure q[1, 3, 5]` as a single node instead of as 3 separate nodes, in terms of moving instructions around, dependencies with other instructions and so on...).

@arcturusannamalai Is the paragraph above clear to you? In OpenSquirrel, we can build a circuit 1) using the CircuitBuilder, and 2) from a cQASM string. For both paths, we end up creating the same types (nodes in an IR, Intermediate Representation, structure). In this case a Measure instruction type. At the moment, when we build a Measure instruction from a cQASM string, and that instruction uses SGMQ (Single Gate Multiple Qubit) notation, we create as many instructions as qubits are used. For the CircuitBuilder path, we don't have the option of using SGMQ (that's what this issue would like to add). But the Measure IR node should remain the same (a type receiving a single qubit).

arcturusannamalai:
So you need "Allow measure all functionality when user does not specify a qubit, i.e Measure() performs a measure all". We can add that perhaps. But seems like there is more requirements beyond that.

rturrado:
> So you need "Allow measure all functionality when user does not specify a qubit, i.e Measure() performs a measure all". We can add that perhaps. But seems like there is more requirements beyond that.

We will add a measure instruction without arguments at some point. But the problem with this PR is other, in my opinion. I don't think a Measure IR node should have multiple qubits.

PR #319, opened 2024-09-06
[CQT 126] Verify dataclasses using __post_init__

juanboschero:


opensquirrel/ir.py
```diff
@@ -76,6 +76,13 @@ class Float(Expression):
     def accept(self, visitor: IRVisitor) -> Any:
         return visitor.visit_float(self)
 
+    def __post_init__(self) -> None:
+        try:
+            self.value = float(self.value)
+        except ValueError as exception:
+            msg = "value must be a float"
+            raise ValueError(msg) from exception
+
```
S-Linde, line 85: I would prefer the following:
```python
def __post_init__(self) -> None:
    if isinstance(self.value, SupportsFloat):
        self.value = float(self.value)
    else:
        msg = "value must be a float"
        raise TypeError(msg)
```
Note that this means that you need to add
```python
from typing import SupportsFloat
```
at the top of the file. (this just checks if `self.value` has a `__float__` method, which is used by `float`.)

Also, the error should be a `TypeError`.

opensquirrel/ir.py
```diff
@@ -84,6 +91,13 @@ class Int(Expression):
     def accept(self, visitor: IRVisitor) -> Any:
         return visitor.visit_int(self)
 
+    def __post_init__(self) -> None:
+        try:
+            self.value = int(self.value)
+        except ValueError as exception:
+            msg = "value must be an int"
+            raise ValueError(msg) from exception
+
```
S-Linde, line 100: See comment above but with `SupportsInt`.

opensquirrel/ir.py
```diff
@@ -95,6 +109,13 @@ def __hash__(self) -> int:
     def __repr__(self) -> str:
         return f"Bit[{self.index}]"
 
+    def __post_init__(self) -> None:
+        try:
+            self.index = int(self.index)
+        except ValueError as exception:
+            msg = "index must be an int"
+            raise ValueError(msg) from exception
```
S-Linde, line 117: Same as above

opensquirrel/ir.py
```diff
@@ -109,6 +130,13 @@ def __hash__(self) -> int:
     def __repr__(self) -> str:
         return f"Qubit[{self.index}]"
 
+    def __post_init__(self) -> None:
+        try:
+            self.index = int(self.index)
+        except ValueError as exception:
+            msg = "index must be an int"
+            raise ValueError(msg) from exception
```
S-Linde, line 138: Same as above

test/test_ir.py
```diff
@@ -297,3 +308,45 @@ class TestControlledGate:
     def test_control_gate_same_control_and_target_qubit(self) -> None:
         with pytest.raises(ValueError, match="control and target qubit cannot be the same"):
             ControlledGate(Qubit(0), BlochSphereRotation(Qubit(0), [0, 0, 1], angle=np.pi))
+
+
+class TestDataClasses:
```
S-Linde, line 313: I would prefer `TestInt`, `TestQubit`, etc.

S-Linde (commented):
I would prefer a slightly different approach for the type checking.

This also raises the question of `Int.__int__` and `Float.__float__` should exists. Or... if `Int` and `Float` should exist at all.

juanboschero:
> I would prefer a slightly different approach for the type checking.
> 
> This also raises the question of `Int.__int__` and `Float.__float__` should exists. Or... if `Int` and `Float` should exist at all.

This is indeed a conversation we need to have, an immediate example of this is [here](https://qutech-sd.atlassian.net/jira/software/c/projects/CQT/boards/18/backlog?selectedIssue=CQT-183).

S-Linde (commented):
I think there are some typos in the code, double check to be sure.

I put a lot of comments around naming and simplifications in the tests, but they are all rather similar. You can use it as a checklist.

S-Linde (approved):

PR #324, opened 2024-09-11
Use SupportsInt instead of Int in type hints

S-Linde:


opensquirrel/circuit_builder.py
```diff
@@ -122,15 +122,14 @@ def _check_generator_f_args(
...
-                if args[i].__class__.__name__ != par.annotation:
-                    msg = (
-                        f"wrong argument type for instruction `{attr}`, "
-                        f"got {type(args[i])} but expected {par.annotation}"
-                    )
-                    raise TypeError(msg)
-            elif not isinstance(args[i], par.annotation):
-                msg = f"wrong argument type for instruction `{attr}`, got {type(args[i])} but expected {par.annotation}"
+            try:
+                expected_type = ANNOTATIONS[par.annotation] if isinstance(par.annotation, str) else par.annotation
```
juanboschero, line 126: This is a very good idea!

juanboschero (approved):

PR #325, opened 2024-09-11
[CQT-127] Allow Qubit to accept QubitLike

juanboschero:
* Make QubitLike type

* Reformat Qubit dataclass

---------


opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -39,13 +38,15 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
             # Rxy rotation.
             theta = round(math.degrees(g.angle), FIXED_POINT_DEG_PRECISION)
             phi: float = round(math.degrees(math.atan2(g.axis[1], g.axis[0])), FIXED_POINT_DEG_PRECISION)
-            self.schedule.add(quantify_scheduler_gates.Rxy(theta=theta, phi=phi, qubit=self._get_qubit_string(g.qubit)))
+            self.schedule.add(
+                quantify_scheduler_gates.Rxy(theta=theta, phi=phi, qubit=self._get_qubit_string(Qubit(g.qubit)))
```
rturrado, line 42: Can we change `_get_qubit_string` to use a `QubitLike` parameter? So that we don't need the conversion to `Qubit` here either.

opensquirrel/ir.py
```diff
@@ -140,23 +140,38 @@ def accept(self, visitor: IRVisitor) -> Any:
...
+
+        Args:
+            index: index of the ``Qubit`` object.
+        """
+        if isinstance(index, SupportsInt):
+            self.index = int(index)
+        elif isinstance(index, Qubit):
+            self.index = index.index
+        else:
+            msg = "index must be an int"
```
rturrado, line 164: Maybe `index must be an instance of SupportsInt or Qubit`?

opensquirrel/ir.py
```diff
@@ -140,23 +140,38 @@ def accept(self, visitor: IRVisitor) -> Any:
...
 class Qubit(Expression):
+    """``Qubit`` is used for intermediate representation of ``Statement`` arguments.
+
+    Attributes:
+        index: index of the ``Qubit`` object.
+    """
+
     index: int
 
+    def __init__(self, index: QubitLike) -> None:
```
rturrado, line 153: Maybe `value` a better name than `index` here.

opensquirrel/utils/matrix_expander.py
```diff
@@ -41,12 +51,13 @@ def get_reduced_ket(ket: int, qubits: Iterable[Qubit]) -> int:
     """
     reduced_ket = 0
     for i, qubit in enumerate(qubits):
+        qubit = Qubit(qubit)
         reduced_ket |= ((ket & (1 << qubit.index)) >> qubit.index) << i
```
rturrado, line 55: Maybe `&` and `<<` and `>>` operations can be overloaded for `Qubit`, so you don't need the line `qubit = Qubit(qubit)`.

I.e., you could be able to use a qubit as an int.

opensquirrel/writer/writer.py
```diff
@@ -59,10 +59,12 @@ def visit_gate(self, gate: Gate) -> None:
                 gate_name = gate_name.replace("\n", "")
             self.output += f"{gate_name}\n"
             return
-        if any(not isinstance(arg, Qubit) for arg in gate.arguments):  # type: ignore[union-attr]
-            params = [arg.accept(self) for arg in gate.arguments if not isinstance(arg, Qubit)]  # type: ignore[union-attr]
+        if any(not isinstance(arg, QubitLike.__args__) for arg in gate.arguments):  # type: ignore
```
rturrado, line 62: This `QubitLike.__args__` looks ugly. Code smell maybe?

test/docs/test_tutorial.py
```diff
@@ -96,19 +95,19 @@ def test_circuit_builder_qft() -> None:
...
+CRk q[3], q[1], q[3]
+CRk q[4], q[1], q[4]
 H q[2]
-CRk(2) q[3], q[2]
-CRk(3) q[4], q[2]
+CRk q[3], q[2], q[2]
+CRk q[4], q[2], q[3]
 H q[3]
-CRk(2) q[4], q[3]
+CRk q[4], q[3], q[2]
```
rturrado, line 110: What happened here? CRk is a gate with an angle param, of float type, and two qubit operands.

rturrado (commented):
Beautiful work!

I've added a few comments and one check to review.

I wouldn't be sure to _always_ stop using a `Qubit(index)` syntax. Sometimes it may be clearer to know the type of the arguments.

opensquirrel/writer/writer.py
juanboschero: If we take this out, python 3.9 crashes `TypeError: Subscripted generics cannot be used with class and instance checks`

juanboschero: This is a bug that was fixed on after but since we support 3.9, we are forced to use `__args__`

opensquirrel/utils/matrix_expander.py
juanboschero: I will make a ticket for this as I think this is out of scope and would require us to investigate whether we would want this functionality or not

opensquirrel/ir.py
juanboschero: I disagree, if I think of a qubit value, I immediately think of the ket state.

rturrado: Fair enough. I thought of `qubit`, but the `__init__` parameter is actually an index value. I only wanted to avoid the `index.index`, which sounded a bit weird to me.

opensquirrel/writer/writer.py
rturrado: Fair enough, thanks.

opensquirrel/utils/matrix_expander.py
juanboschero: It is usually good practice in Python to allow your function inputs to be as broad as possible and your return to be as specific as possible. I fear that by doing this, we would force Qubit objects onto `get_reduced_ket`. 

I think a ticket is good, this is not the only function where this functionality should be instilled. I think this conversion should be part of the IR itself.

rturrado: > It is usually good practice in Python to allow your function inputs to be as broad as possible and your return to be as specific as possible.
That can make sense, but not in all cases. I think this is one of those cases when you want to be more specific.

> I fear that by doing this, we would force Qubit objects onto `get_reduced_ket`.
That's it, but I think that's a good thing in this case.

> I think a ticket is good, this is not the only function where this functionality should be instilled.
I think that's going too far, really.

opensquirrel/ir.py
```diff
@@ -140,23 +140,38 @@ def accept(self, visitor: IRVisitor) -> Any:
...
+
+        Args:
+            index: index of the ``Qubit`` object.
+        """
+        if isinstance(index, SupportsInt):
+            self.index = int(index)
+        elif isinstance(index, Qubit):
+            self.index = index.index
+        else:
+            msg = "index must be an instance of SupportsInt or Qubit"
```
rturrado, line 164:
```suggestion
            msg = "index must be a QubitLike"
```

opensquirrel/circuit_builder.py
```diff
@@ -128,11 +129,18 @@ def _check_generator_f_args(
...
+            except TypeError:
+                # expected type is probably a Union, which works differently in python39
+                is_incorrect_type = not isinstance(args[i], expected_type.__args__)
+
+            if is_incorrect_type:
                 msg = f"wrong argument type for instruction `{attr}`, got {type(args[i])} but expected {expected_type}"
                 raise TypeError(msg)
-            if args[i].__class__.__name__ == "Qubit":
-                self._check_qubit_out_of_bounds_access(args[i].index)
+            if expected_type in (QubitLike, Qubit):
```
rturrado, line 142: Can this be simplified to `expected_type in QubitLike`?

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -17,14 +17,13 @@
...
 
-
 # Radian to degree conversion outcome precision
 FIXED_POINT_DEG_PRECISION = 5
 
 
 class _ScheduleCreator(IRVisitor):
-    def _get_qubit_string(self, q: Qubit) -> str:
-        return f"{self.qubit_register_name}[{q.index}]"
+    def _get_qubit_string(self, q: QubitLike) -> str:
```
rturrado, line 25: I would say this is yet another case where we want to restrict `q` to being a `Qubit`. This is an internal function. If you have a look at the points where this function is called, you'll see we always call it with a qubit.

So we can get rid of all the `Qubit()` conversions.

rturrado (commented):
I have done a more in depth review now.

My main comment is:
- I don't think we have to change to using `QubitLike` everywhere. For example, for internal functions, such as `expand_ket`, that we are alredy calling with operands of type `Qubit`.
- It's enough letting the programer use integers instead of `Qubit` types in the code that is, let's say, exposed. For example, through the CircuitBuilder, when you construct a gate.

My other bigger concern was: can't we have an implicit "from int" constructor? But I see constructor overloading is not really a built-in Python feature.

Other than that, it looks great to me.

opensquirrel/circuit_builder.py
juanboschero: This does not work because `TypeError: argument of type '_UnionGenericAlias' is not iterable` 

Placing .__args__ introduces regex issues in the writer.

rturrado: Cool. Thanks for checking!

opensquirrel/ir.py
```diff
@@ -656,14 +681,18 @@ def accept(self, visitor: IRVisitor) -> None:
             statement.accept(visitor)
 
 
+# Type Aliases
+AxisLike = Union[ArrayLike, Axis]
+QubitLike = Union[SupportsInt, Qubit]
+
+
 ANNOTATIONS = {
```
rturrado, line 689: I know this is not part of this PR. But any idea why `ANNOTATIONS` is spelled all in capitals? Is it because it's a constant dictionary? Is this a proper name also? Maybe `typename_to_type_map` as a suggestion? Or  `TYPENAME_TO_TYPE_MAP`?

rturrado: But, as I commented elsewhere, I would consider turning it into a class with class methods, like `is_qubit_like(str) -> bool`.

opensquirrel/ir.py
```diff
@@ -535,6 +552,12 @@ def named_gate(gate_generator: Callable[..., ControlledGate]) -> Callable[..., C
 
 
 def named_gate(gate_generator: Callable[..., Gate]) -> Callable[..., Gate]:
+    def is_qubit_like_annotation(annotation: Any) -> bool:
+        return annotation in (QubitLike, Qubit)
+
+    def is_supports_int_annotation(annotation: Any) -> bool:
```
rturrado, line 558: Can we?:

1. Put these functions close to `ANNOTATIONS_TO_TYPE_MAP` .
2. Try to use them anywhere else in the code, at least in this `ir.py` file? There is more code like `annotation in (blah, foo)` around.

Thanks!

opensquirrel/utils/matrix_expander.py
```diff
@@ -8,10 +8,20 @@
...
+    ControlledGate,
+    Gate,
+    IRVisitor,
+    MatrixGate,
+    Qubit,
+    QubitLike,
+)
+
+
+def get_reduced_ket(ket: int, qubits: Iterable[QubitLike]) -> int:
```
rturrado, line 24: Can we change `QubitLike` to `Qubit` here? As it was done for `expand_ket`.

rturrado (commented):
Some last comments.

Thanks for the changes.

opensquirrel/ir.py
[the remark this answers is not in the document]
juanboschero: Nice catch!

rturrado (approved):
I think all conversations are resolved now, so let's go!

PR #349, opened 2024-10-03
CQT-141-Qubit-declaration-added-to-simplest-circuit

GuyPuts:
Adapts the Writer class such that empty circuit declarations (which are the default for a qubit register in the backend) are not printed to console to avoid confusion.


opensquirrel/writer/writer.py
```diff
@@ -13,10 +13,11 @@ def __init__(self, register_manager: RegisterManager) -> None:
...
         qubit_register_size = self.register_manager.get_qubit_register_size()
         qubit_register_name = self.register_manager.get_qubit_register_name()
+
         bit_register_size = self.register_manager.get_bit_register_size()
         bit_register_name = self.register_manager.get_bit_register_name()
-        self.output = "version 3.0\n\n{}\n{}\n".format(
-            f"qubit[{qubit_register_size}] {qubit_register_name}",
+        self.output = "version 3.0\n".format(
+            "\n{}\n{}\n"+f"qubit[{qubit_register_size}] {qubit_register_name}" if qubit_register_size > 0 else "",
             f"bit[{bit_register_size}] {bit_register_name}\n" if bit_register_size > 0 else "",
```
rturrado, line 21:
```suggestion
        self.output = "version 3.0{}{}{}{}\n".format(
            "\n\n" if qubit_register_size > 0 or bit_register_size > 0,
            f"qubit[{qubit_register_size}] {qubit_register_name}" if qubit_register_size > 0 else "",
            "\n" if qubit_register_size > 0 and bit_register_size > 0 else "",
            f"bit[{bit_register_size}] {bit_register_name}" if bit_register_size > 0 else "",
```

I would suggest this. A `version 3.0` line followed by 3 fields:
1. If we only have qubits, we'll write: version, blank line, qubits.
2. If we qubits and bits, we'll write: version, blank line, qubits, bits.
3. If we only have bits, we'll write: version, blank line, bits.
4. If we don't have neither qubits nor bits, we'll write: version.

All the options end with a newline.

rturrado (commented):
Good work Guy!

Can we add some tests for this code? Circuits:

- With qubits and bits.
- Only with qubits.
- Only with bits.
- Without qubits and bits.

Thanks!

opensquirrel/writer/writer.py
GuyPuts: Hi Robert, what would be the best place to test this? I would say test_integration.py, or do you think another place would be better?

rturrado: I would probably add the tests to `test/parser/libqasm/test_libqasm.py`. There you have already a `test_simple`. You could add, for example, `test_simplest_qubits_but_no_bits`, `test_simplest_qubits_and_bits`, `test_simplest_no_qubits_but_bits`, and `test_simplest_no_qubits_and_no_bits`.

rturrado (changes requested):
Let's refine those tests, Guy!

test/writer/test_writer.py
```diff
@@ -133,3 +133,69 @@ def test_comment() -> None:
...
+qubit[1] q
+"""
+    )
+    # Qubit and bit
+    qc = Circuit.from_string(
+        """
+        version 3.0
+
+        qubit[1] q
+        bit[1] b
```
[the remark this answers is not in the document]
GuyPuts: Solved in the next commit :)

test/writer/test_writer.py
```diff
@@ -133,3 +133,69 @@ def test_comment() -> None:
...
+
+def test_string_representation() -> None:
+    """Tests the string representation of a circuit."""
+    # Only qubit
+    qc = Circuit.from_string(
+        """
+        version 3.0
+
+        qubit[1] q
+
```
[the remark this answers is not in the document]
GuyPuts: Solved in the next commit :)

test/writer/test_writer.py
```diff
@@ -133,3 +133,69 @@ def test_comment() -> None:
 CR(1.234) q[0], q[1]
 """
     )
+
+
+def test_string_representation() -> None:
```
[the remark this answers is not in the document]
GuyPuts: Yes indeed

tox.ini
```diff
@@ -15,6 +15,7 @@ commands_pre =
 description = run linters
 commands =
     poetry run ruff check
+    poetry run ruff format --check
```
rturrado, line 18: @elenbaasc @S-Linde are you OK with this change? Is this a joint decision?

test/writer/test_writer.py
```diff
@@ -129,60 +129,3 @@ def test_comment() -> None:
 CR(1.234) q[0], q[1]
 """
     )
-
```
rturrado: I don't understand why this change is appearing in the code review, because this test was added and then removed by this pull request.

test/parser/libqasm/test_libqasm.py
```diff
@@ -78,3 +109,56 @@ def test_error(parser: Parser) -> None:
 def test_wrong_gate_argument_number_or_types(parser: Parser, error_message: str, circuit_string: str) -> None:
     with pytest.raises(IOError, match=error_message):
         parser.circuit_from_string(circuit_string)
+
+
+@pytest.mark.parametrize(
+    ("expected_output", "circuit_string"),
```
rturrado, line 115: In order to be consistent with other parametrized tests across the baseline, let's put `expected_output` as second element.

test/parser/libqasm/test_libqasm.py
```diff
@@ -78,3 +109,56 @@ def test_error(parser: Parser) -> None:
...
+            """version 3.0
+
+qubit[1] q
+""",
+            """
+            version 3.0
+
+            qubit q
+
+            """,
```
rturrado, line 127: Can we write this as `"version 3.0; qubit q"`? Notice no newline at the end.

test/parser/libqasm/test_libqasm.py
```diff
@@ -78,3 +109,56 @@ def test_error(parser: Parser) -> None:
...
+
+
+@pytest.mark.parametrize(
+    ("expected_output", "circuit_string"),
+    [
+        (
+            """version 3.0
+
+qubit[1] q
+""",
```
rturrado, line 121: Can we write this as `"version 3.0\n\nqubit[1] q"`? Just to simplify the test.

test/parser/libqasm/test_libqasm.py
```diff
@@ -78,3 +109,56 @@ def test_error(parser: Parser) -> None:
...
+            """
+            version 3.0
+
+            bit b
+
+            """,
+        ),
+        (
+            """version 3.0\n""",
+            """version 3.0""",
```
rturrado, line 157: Do we need the triple double quotes? Or can we just write double quotes?

test/parser/libqasm/test_libqasm.py
```diff
@@ -78,3 +109,56 @@ def test_error(parser: Parser) -> None:
...
+            """version 3.0
+
+bit[1] b
+""",
+            """
+            version 3.0
+
+            bit b
+
+            """,
```
rturrado, line 153: Same comments as above. Can we write all this in single lines? Also, do no use unnecessary blank lines at the end of input programs.

rturrado (commented):
Thanks for the changes Guy.

Just a question and a comment:
- Question: have we agreed on adding the `ruff format` to tox.ini?
- Comment: can you try to write the circuit strings and the expected outputs in a single line? Just to make the test simpler. You could even put a circuit string and an expected output in the same line.

Thanks!

test/parser/libqasm/test_libqasm.py
GuyPuts: I added the test you mentioned, as to test both ```version 3.0; qubit q``` and ```version 3.0 \n qubit q```

tox.ini
GuyPuts: I added this line per Stan's suggestion. @S-Linde

test/parser/libqasm/test_libqasm.py
GuyPuts: Yes I will change it :)

tox.ini
GuyPuts: @S-Linde @rturrado I added this line to the tox.ini, but the pipelines are still failing on the formatting. Do you know what might be the issue in this case?

test/parser/libqasm/test_libqasm.py
```diff
@@ -46,35 +64,69 @@ def test_sgmq(parser: Parser) -> None:
...
+        ),
+    ],
+    ids=[
+        "only_qubit",
+        "qubit_and_bit",
+        "only_bit",
+        "empty_declaration",
+        "semicolon_qubit",
+        "semicolon_qubit_and_bit",
+        "semicolon_bit",
```
rturrado, line 127: Beautiful! And thanks for adding the semicolon versions ;-)

rturrado (approved):

tox.ini
rturrado: It seems 2 out of 80 files need reformatting. Try running `poetry run ruff check --fix` and `poetry run ruff format` locally.

GuyPuts: I ran the commands with the changes to tox.ini and it works now :)

rturrado (approved):

PR #357, opened 2024-10-15
Fix Numpy at < 2.0 for Python 3.9

elenbaasc:
- Adjusted pyproject.toml file to fix Numpy ^1.26 for Python 3.9
- Update poetry.lock file


Mythir (approved):

juanboschero (approved):
Unfortunately, this is the only way to allow py3.9 to work with isinstance of a union

PR #363, opened 2024-10-18
Remove unsupported gates from default gate set.

elenbaasc:
- Removed SWAP, sqrtSWAP, and CCZ from the `default_gates.py`
- Removed tests specifically involving these gates

### DIFF
--- a/opensquirrel/default_gates.py
+++ b/opensquirrel/default_gates.py
@@ -4,9 +4,7 @@ import math
 from collections.abc import Callable
 from typing import SupportsInt
 
-import numpy as np
-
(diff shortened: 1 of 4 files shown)


rturrado:
It may be useful tagging `develop` right after merging this so that, when we have to add this again in the future, we quickly localize the commits to _reverse cherry-pick_.

```
git tag just_after_removing_swap_sqrt-swap_and_ccz
git push origin just_after_removing_swap_sqrt-swap_and_ccz
```

But I would ask @oschusler first to see if this is a good practice or there may be better ways (this way, I suppose we can quickly _dirty_ the list of tags).

rturrado (approved):

PR #367, opened 2024-10-24
[CQT-235] Add support for gate modifiers

rturrado:
## Notes

- This is very much a work in progress.
- The current PR only passes `test_gate_modifiers` and `test_named_gates`, at `test_libqasm.py`.

## Goal

- This PR updates OpenSquirrel to use libQASM 0.6.8.
- libQASM 0.6.8 introduces 2 main changes:
1. Gate modifiers feature is implemented.
2. Parameters are no longer part of an instruction signature.

The semantic AST that is now passed onto OpenSquirrel has this form:

- A statement can be a variable declaration, a gate instruction, or a non-gate instruction.
- Gate instructions correspond to unitary operations, i.e., either (named) gates (such as `X`), or combinations of gate modifiers acting on a (named) gate (such as `ctrl.pow(2).inv.X`).
- Non-gate instructions correspond to non-unitary operations, i.e., `measure` or `reset`.
- Gate instructions have a `Gate` and a list of operands. The operands are of `Qubit` type.
- A `Gate` can have, optionally, another `Gate`, and also optionally, a parameter.
  Examples:
  - For `inv.X`, `inv` is a gate (modifier) that has an `X` gate and no parameter.
  - For `pow(2).Y`, `pow` is a gate (modifier) that has a `Y` gate and a `2` parameter (an exponent of type float).
  - Similarly, `X` has neither gate nor parameter, and `Rx` has no gate but a parameter (an angle of type float).

## Changes

In order to address the 2 main changes mentioned above, the current implementation of this PR has experimented with the following changes:
1. Implement gate modiiers as functors. They are implemented in `default_gate_modifiers.py`. They are called `{Inverse,Power,Control}GateModifier`.
2. Reimplement (named) gates as functors.

The idea behind this change from functions to functors is, basically, not to treat parameters as arguments but as members of a given instruction.
For example, for `Rx`, we won't be calling anymore `Rx(qubit: Qubit, theta: Float)`. Instead, we will have an `Rx(theta: Float)` object, and we will call `Rx(qubit)`.
The same applies to `PowerGateModifier`.

## TODO

As commented in the Notes above, the current PR has only been tested with a couple of parametrized tests.

1. Check what errors arise from running the rest of the tests.
2. Check if the reimplementation of (named) gates as functors causes major disruptions in the rest of the code.
  It could be that, in many parts of the code, we are just using a simple syntax such as `X(q0)`, and this needs now a syntax `X()(q0)`, i.e., instantiate the `X` functor and then invoke `X.__call__(q0)`.
  In case it does, we should investigate if the current implementation can be improved so that a syntax such as `X(q0)` still works with functors.
  Or, even, we may rethink the reimplementation to not use functors.


opensquirrel/default_gates.py
```diff
@@ -1,117 +1,211 @@
...
+    def __call__(self, control: QubitLike, target: QubitLike) -> ControlledGate:
+        self.theta = 2 * math.pi / (2 ** Int(self.parameter).value)
+        return ControlledGate(
+            control,
+            BlochSphereRotation(qubit=target, axis=(0, 0, 1), angle=self.theta, phase=self.theta / 2)
+        )
 
 
-default_bloch_sphere_rotations_without_params: list[Callable[[QubitLike], BlochSphereRotation]]
+default_bloch_sphere_rotations_without_params: list[type[NamedGateFunctor]]
```
juanboschero, line 208: Would this not also include multi-qubit gates?

opensquirrel/default_gates.py
```diff
@@ -1,117 +1,211 @@
...
+
+    def __call__(self, q: QubitLike) -> BlochSphereRotation:
+        self.theta = Float(self.parameter).value
+        return BlochSphereRotation(qubit=q, axis=(0, 0, 1), angle=self.theta, phase=0)
 
 
 @named_gate
-def CNOT(control: QubitLike, target: QubitLike) -> ControlledGate:
-    return ControlledGate(control, X(target))
+class CNOT(NamedGateFunctor):
```
juanboschero, line 165: If gates are now classes, would the matrix representation of the gate be contained in the class itself?

opensquirrel/default_gates.py
```diff
@@ -1,117 +1,211 @@
...
+
+    def __call__(self, q: QubitLike) -> BlochSphereRotation:
+        self.theta = Float(self.parameter).value
+        return BlochSphereRotation(qubit=q, axis=(1, 0, 0), angle=self.theta, phase=0)
 
 
 @named_gate
-def Ry(q: QubitLike, theta: Float) -> BlochSphereRotation:
-    return BlochSphereRotation(qubit=q, axis=(0, 1, 0), angle=theta.value, phase=0)
+class Ry(NamedGateFunctor):
```
juanboschero, line 145: Will gates possess __hash__, __eq__, __contains___ etc?

juanboschero (commented):
I like the idea so far! Would be good to brainstorm the points I have left below

opensquirrel/default_gates.py
rturrado: It could.

opensquirrel/default_gates.py
rturrado: If needed, I don't see why not.

opensquirrel/default_gates.py
rturrado: Good point. We may still have `BlochSphereRotation`s as subclasses of `NamedGateFunctor`s.

PR #368, opened 2024-10-28
[CQT-43] Take global phases into account when verifying decompositions

juanboschero:
- Changed decomposer object logic
- Phase of qubits are now tracked by circuit object
- Tests added to verify and correct global phase leak into relative phase


test/test_integration.py
```diff
@@ -115,6 +115,81 @@ def test_Spin2_backend() -> None:
...
+def test_integration_global_phase() -> None:
+    qc = Circuit.from_string(
+        """
+        version 3.0
+
+        // This is a single line comment which ends on the newline.
+        // The cQASM string must begin with the version instruction (apart from any preceding comments).
+
+        /* This is a multi-
+        line comment block */
```
rturrado, line 127:
```suggestion
```

test/test_integration.py
```diff
@@ -115,6 +115,81 @@ def test_Spin2_backend() -> None:
     )
 
 
+def test_integration_global_phase() -> None:
```
rturrado, line 118: Can we add a comment somewhere in the code saying where a global phase is expected to be introduced?

test/test_integration.py
```diff
@@ -115,6 +115,81 @@ def test_Spin2_backend() -> None:
...
+X90 q[0]
+Rz(3.1415927) q[0]
+CZ q[2], q[0]
+Rz(1.5707963) q[0]
+X90 q[0]
+Rz(1.5707963) q[0]
+Rz(3.1415927) q[1]
+X90 q[1]
+Rz(2.8235927) q[1]
+X90 q[1]
```
rturrado, line 188: Just out of curiosity, what instructions, if any, do the global phase correction here?

rturrado (changes requested):
Nice work!

I would mainly have a look at the check_gate_replacements function.

I would also try to:
- Encapsulate the phase map in a class.
- Use more functions.
- Make the code more readable.

opensquirrel/circuit.py
```diff
@@ -166,3 +149,31 @@ def export(self, fmt: ExportFormat | None = None) -> Any:
...
+    def __init__(self, phase_map: ArrayLike[np.complex128]):
+        """Initialize a PhaseMap object."""
+        self.phase_map = phase_map
+
+    def __contains__(self, qubit: QubitLike) -> bool:
+        """Checks if qubit is in the phase map."""
+        if qubit in self.phase_map:
+            return True
+
+        return False
```
rturrado, line 166:
```suggestion
        return qubit in self.phase_map
```

opensquirrel/circuit.py
```diff
@@ -166,3 +149,31 @@ def export(self, fmt: ExportFormat | None = None) -> Any:
...
+
+    def set_phase_map(self, phase_map: NDArray[np.complex128]) -> None:
+        self.qubit_phase_map = phase_map
+
```
rturrado, line 170:
```suggestion
```

I think this is a copy construction, and should be taken care of in the/a constructor.

rturrado (changes requested):

opensquirrel/decomposer/general_decomposer.py
```diff
@@ -32,25 +55,37 @@ def check_gate_replacement(gate: Gate, replacement_gates: Iterable[Gate]) -> Non
         msg = f"replacement for gate {gate.name} does not preserve the quantum state"
         raise ValueError(msg)
 
+    if qc is not None:
```
elenbaasc, line 58: I guess `qc` should be `circuit`, but then the methods that are called on `circuit` are the methods of the `PhaseMap` class, i.e., not of the Circuit class.

I'm still confused what the code block (of this if-statement) actually does. I get the idea, but I feel that this might actually _introduce_ relative phases, instead of mitigating them...

Can you better explain in the Confluence page what is being implemented and how that prevents a global phase difference in the gates (between original and replacement), from resulting in a relative phase between the qubits, because that bit is still unclear to me. To put it another way: How can a global phase difference in a gate replacement, lead to a relative phase difference. As far as I see it, this can only happen if a replaced gate (that differs by a global phase from the original) is made into a controlled gate, but this doesn't happen anywhere in the compilation process (yet).

elenbaasc (commented):
This still needs some looking into.

I think the phase map may be useful (necessary), but I'm not convinced yet.

opensquirrel/circuit.py
```diff
@@ -132,14 +136,44 @@ def replace(self, gate_generator: Callable[..., Gate], f: Callable[..., list[Gat
...
         if fmt == ExportFormat.CQASM_V1:
             from opensquirrel.exporter import cqasmv1_exporter
+
             return cqasmv1_exporter.export(self)
         msg = "unknown exporter format"
         raise ValueError(msg)
+
+
+@dataclass(init=False)
+class PhaseMap(Circuit):
```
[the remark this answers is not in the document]
rturrado: It's not that it's not clear. It's wrong. It's a wrong way to get access to information from the base class. But inheritance should only be used when there is an *is a* relationship with the base class.

A PhaseMap should be an independent data structure, managing some data, e.g., a map of phases, encapsulating the definition of that map of phases (today a dictionary, tomorrow whatever), and providing a given API to access, query, etc. that data.

If PhaseMap turns out to be very simple, e.g., just a map over which we only do set and get, we can decide not to use it. But at the very first moment that it is more complex than that, you won't want to be writing complicating code all around the code base to access a map of phases. In that case, we'd better have a PhaseMap.

PR #369, opened 2024-10-30
Cqt 115 pep8 naming

GuyPuts:
Implements PEP8 styling convention, except for function names

### DIFF
--- a/opensquirrel/exporter/quantify_scheduler_exporter.py
+++ b/opensquirrel/exporter/quantify_scheduler_exporter.py
@@ -7,7 +7,15 @@ from opensquirrel.circuit import Circuit
 from opensquirrel.common import ATOL
 from opensquirrel.default_gates import X, Z
 from opensquirrel.exceptions import ExporterError, UnsupportedGateError
-from opensquirrel.ir import BlochSphereRotation, ControlledGate, IRVisitor, MatrixGate, Measure, Qubit, Reset
+from opensquirrel.ir import (
+    BlochSphereRotation,
+    ControlledGate,
+    IRVisitor,
+    MatrixGate,
+    Measure,
+    Qubit,
+    Reset,
+)
 
 try:
     import quantify_scheduler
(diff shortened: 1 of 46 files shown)


elenbaasc:
Hi Guy, thanks for looking at this.

Note that the pep8-naming option needs to be uncommented in the options of `[tool.ruff.lint]` in `pyproject.toml`, and then ruff needs to be run again (`poetry run ruff check`) to see if everything conforms to the pep8 naming standard.

To ignore pep8 naming (for instance for the gate names)  you can add the `noqa` (no quality assurance) directive followed by the type (code) of ruff error to ignore, e.g.,
`def H(q: QubitLike) -> BlochSphereRotation:  # noqa: N802`
Note that there are two spaces before the `#` and one space after (otherwise ruff format will complain) and that you need to  run ruff to determine the codes (above is N802, to ignore _Function names that should be lowercase_.)

juanboschero (dismissed):

opensquirrel/exporter/quantify_scheduler_exporter.py
```diff
@@ -40,7 +48,8 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
         # Note that when adding a rotation gate to the Quantify-scheduler Schedule,
         # there exists an ambiguity with how Quantify-scheduler will store an angle of 180 degrees.
         # Depending on the system the angle may be stored as either 180 or -180 degrees.
-        # This ambiguity has no physical consequences, but may cause the exporter test fail.
+        # This ambiguity has no physical consequences, but may cause the
+        # exporter test fail.
```
juanboschero, line 52:
```suggestion
        # exporter test to fail.
```

juanboschero (changes requested):

opensquirrel/exporter/quantify_scheduler_exporter.py
GuyPuts: This has been fixed :)

juanboschero (approved):

PR #372, opened 2024-11-07
[CQT-151] Move compilation passes to passes folder

GuyPuts:
Moves the compilation passes functionalities to the passes folder. CQT-151.


rturrado (commented):
Beautiful!

My only question would be the `utils/check_passes` name. I would usually go for sustantives more than verbs for folder names, e.g., `utils/checks` or `utils/passes` (even `utils/passes_checks`, but I don't really like that one). We could even leave it like `utils/check_mapper.py` until we have more things within `utils`.

Thanks!

rturrado:
> I agree with @rturrado to use nouns (sustantivos in Spanish 😅) for folder names over verbs. The `check_mapper.py` does not need to be in a separate folder, i.e., it can just be in `utils`.

That's what I meant, yes: substantives, not sustantives.

rturrado (changes requested):
@GuyPuts I think you have an issue in your environment. Your environment is forcing you to use an 80 column width, while we are using a 120 column width in OpenSquirrel (tox.ini). Please fix it.

That "fix conflicts" commit should be reverted.

rturrado (dismissed):
Great! Good work! Thanks.

README.md
```diff
@@ -97,7 +97,7 @@ The different decomposition strategies can be found in the
 In the example below, the circuit is decomposed using the Z-Y-Z decomposer.
 
 ```python
-from opensquirrel.decomposer.aba_decomposer import ZYZDecomposer
+from opensquirrel.passes.decomposer.aba_decomposer import ZYZDecomposer
```
elenbaasc, line 100: Sorry for the last minute comment:

Because of the way the `__init__.py` file in the in the `decomposer` directory is configured, you can import the various decomposers directly from the `decomposer` directory, _e.g._,
```suggestion
from opensquirrel.passes.decomposer import ZYZDecomposer
```

This should work for all decomposers and maybe nice to add this change to this PR still...

elenbaasc: @GuyPuts Do you mind if I make these changes and finalize the PR before this branch becomes too much delayed?

GuyPuts: Picking it up right now. This does not work for the general decomposer, by the way

rturrado (approved):
Beautiful!

PR #373, opened 2024-11-08
Fix PEP8 naming

rturrado:
Reverted unwanted changes from [CQT-115] PEP8 naming.
This PR still keeps PEP8 naming enabled in pyproject.tml.

### DIFF
--- a/opensquirrel/decomposer/cnot_decomposer.py
+++ b/opensquirrel/decomposer/cnot_decomposer.py
@@ -37,55 +37,39 @@ class CNOTDecomposer(Decomposer):
         # This gives us an ABC decomposition (U = AXBXC, ABC = I) of the target gate.
         # See https://threeplusone.com/pubs/on_gates.pdf
 
-        # Try special case first, see
-        # https://arxiv.org/pdf/quant-ph/9503016.pdf lemma 5.5
+        # Try special case first, see https://arxiv.org/pdf/quant-ph/9503016.pdf lemma 5.5
         controlled_rotation_times_x = general_merger.compose_bloch_sphere_rotations(X(target_qubit), g.target_gate)
-        (
-            theta0_with_x,
-            theta1_with_x,
-            theta2_with_x,
-        ) = ZYZDecomposer().get_decomposition_angles(
-            controlled_rotation_times_x.angle, controlled_rotation_times_x.axis
+        theta0_with_x, theta1_with_x, theta2_with_x = ZYZDecomposer().get_decomposition_angles(
+            controlled_rotation_times_x.angle,
+            controlled_rotation_times_x.axis,
         )
         if abs((theta0_with_x - theta2_with_x) % (2 * math.pi)) < ATOL:
             # The decomposition can use a single CNOT according to the lemma.
-            A = [  # noqa: N806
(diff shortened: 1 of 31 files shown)


opensquirrel/ir.py
```diff
@@ -270,13 +265,7 @@ def _normalize_axis(axis: NDArray[np.float64]) -> NDArray[np.float64]:
...
         return axis / np.linalg.norm(axis)
 
-    @overload
-    def __getitem__(self, i: int, /) -> np.float64: ...
-
-    @overload
-    def __getitem__(self, s: slice, /) -> list[np.float64]: ...
-
-    def __getitem__(self, index: int | slice, /) -> np.float64 | list[np.float64]:
+    def __getitem__(self, index: int, /) -> np.float64:  # type:ignore[override]
```
elenbaasc, line 268: I agreed to this change. Why are we ignoring this?

opensquirrel/decomposer/cnot_decomposer.py
```diff
@@ -37,55 +37,39 @@ def decompose(self, g: Gate) -> list[Gate]:
...
             # The decomposition can use a single CNOT according to the lemma.
-            A = [  # noqa: N806
-                Ry(target_qubit, Float(-theta1_with_x / 2)),
-                Rz(target_qubit, Float(-theta2_with_x)),
-            ]
-            B = [  # noqa: N806
-                Rz(target_qubit, Float(theta2_with_x)),
-                Ry(target_qubit, Float(theta1_with_x / 2)),
-            ]
+            gate_a = [Ry(target_qubit, Float(-theta1_with_x / 2)), Rz(target_qubit, Float(-theta2_with_x))]
```
elenbaasc, line 48: These are not gates, but lists of gates that are summarized as a single operation. I would like to ignore PEP8 here, so we can call them by how these linear operators are known in the theory, i.e., referring to "This gives us an ABC decomposition (U = AXBXC, ABC = I) of the target gate.".

elenbaasc (commented):
Thanks, looks good! Some small comments.

opensquirrel/ir.py
rturrado: That's a question for @GuyPuts. This went in as a change in CQT-115. If you consider it, we can revert it.

opensquirrel/decomposer/cnot_decomposer.py
rturrado: Fair enough.

opensquirrel/ir.py
elenbaasc: I added it to his PR (with his permission), because he was having difficulties with mypy.

rturrado: So you want to revert this `type:ignore[override]` change?

elenbaasc: Yes, please. That is, if it doesn't fail 😅

rturrado: It looks it doesn-t.

elenbaasc (approved):

rturrado:
So, as far as I see, most of, if not all of, the cases where we override PEP8 naming rules are due to identifiers including an uppercase: A, B, C, H, X, test_Spin2_backend...

PR #378, opened 2024-11-18
[CQT-268] Remove comments from IR

rturrado:
Remove Comment from IR.
Change CircuitBuilder so that it doesn't accept comments.


elenbaasc (approved):
Looks all good :)

PR #379, opened 2024-11-19
Cqt 265 mc kay decomposer issue with specific gate

juanboschero:
- Add tests
- Fix decomposition

### DIFF
--- a/test/decomposer/test_aba_decomposer.py
+++ b/test/decomposer/test_aba_decomposer.py
@@ -1,0 +1,21 @@
+import pytest
+
+from opensquirrel.decomposer import XYXDecomposer
+from opensquirrel.decomposer.general_decomposer import check_gate_replacement
+from opensquirrel.ir import BlochSphereRotation
+
+# Any of the variations of the ABA decomposer has this bug.
+DECOMPOSER = XYXDecomposer
+
+
+@pytest.fixture(name="decomposer")
+def decomposer_fixture() -> DECOMPOSER:
+    return DECOMPOSER()
+
+
+def test_specific_bloch_rotation(decomposer: DECOMPOSER) -> None:
+    axis = [-0.53825, -0.65289, -0.53294]
+    angle = 1.97871
+    arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
+    decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
--- a/test/decomposer/test_aba_decomposer.py
+++ b/test/decomposer/test_aba_decomposer.py
@@ -1,21 +1,35 @@
 import pytest
 
-from opensquirrel.decomposer import XYXDecomposer
+from opensquirrel.decomposer import (
+    XYXDecomposer,
+    XZXDecomposer,
+    YXYDecomposer,
+    YZYDecomposer,
+    ZXZDecomposer,
+    ZYZDecomposer,
+)
 from opensquirrel.decomposer.general_decomposer import check_gate_replacement
 from opensquirrel.ir import BlochSphereRotation
 
-# Any of the variations of the ABA decomposer has this bug.
-DECOMPOSER = XYXDecomposer
+DECOMPOSER = [XYXDecomposer, XZXDecomposer, YXYDecomposer, YZYDecomposer, ZXZDecomposer, ZYZDecomposer]
 
 
-@pytest.fixture(name="decomposer")
-def decomposer_fixture() -> DECOMPOSER:
-    return DECOMPOSER()
-
-
-def test_specific_bloch_rotation(decomposer: DECOMPOSER) -> None:
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
+def test_specific_bloch_rotation(decomposer_class: DECOMPOSER) -> None:
+    decomposer = decomposer_class()
     axis = [-0.53825, -0.65289, -0.53294]
     angle = 1.97871
+
+    arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
+    decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
+def test_negative_bloch_rotation(decomposer_class: DECOMPOSER) -> None:
+    decomposer = decomposer_class()
+    axis = [-0.5789, -0.9000, -3.53294]
+    angle = 3.12
     arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
     decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
     check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
--- a/opensquirrel/decomposer/aba_decomposer.py
+++ b/opensquirrel/decomposer/aba_decomposer.py
@@ -107,6 +107,9 @@ class ABADecomposer(Decomposer, ABC):
         theta1 = (p + m) / 2
         theta3 = p - theta1
 
+        if all(i < 0 for i in axis):
+            return theta3, theta2, theta1
+
         return theta1, theta2, theta3
 
     def decompose(self, g: Gate) -> list[Gate]:


opensquirrel/passes/decomposer/aba_decomposer.py
```diff
@@ -107,6 +107,9 @@ def get_decomposition_angles(self, alpha: float, axis: AxisLike) -> tuple[float,
         theta1 = (p + m) / 2
         theta3 = p - theta1
 
+        if all(i < 0 for i in axis):
+            return theta3, theta2, theta1
```
elenbaasc, line 111:
```suggestion
        if all(component < 0 for component in axis):
            theta1, theta3 = theta3, theta1
```

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,38 @@
...
+from opensquirrel.ir import BlochSphereRotation
+from opensquirrel.passes.decomposer import (
+    Decomposer,
+    XYXDecomposer,
+    XZXDecomposer,
+    YXYDecomposer,
+    YZYDecomposer,
+    ZXZDecomposer,
+    ZYZDecomposer,
+)
```
rturrado, line 14: Can we rewrite this as something like?

```
from opensquirrel.passes.decomposer import Decomposer
from opensquirrel.passes.decomposer import aba_decomposer
```

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,38 @@
...
+    axis = [-0.53825, -0.65289, -0.53294]
+    angle = 1.97871
+
+    arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
+    decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
+def test_negative_bloch_rotation(decomposer_class: Callable[..., Decomposer]) -> None:
```
rturrado, line 32:
```suggestion
def test_axis_with_three_negative_components(decomposer_class: Callable[..., Decomposer]) -> None:
```

rturrado (commented):
Beautiful!

test/decomposer/test_mckay_decomposer.py
```diff
@@ -77,3 +78,16 @@ def test_arbitrary(decomposer: McKayDecomposer) -> None:
         X90(0),
         Rz(0, Float(2.2329420137988887)),
     ]
+
+
+def test_full_sphere(decomposer: McKayDecomposer) -> None:
```
rturrado, line 83:
```suggestion
def test_all_octants_of_bloch_sphere_rotation(decomposer: McKayDecomposer) -> None:
```

test/decomposer/test_mckay_decomposer.py
```diff
@@ -2,6 +2,7 @@
 
 import math
 
+import numpy as np
```
rturrado, line 5: Do we still need the `test_arbitrary` if we have the `test_full_sphere`?

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,44 @@
...
+    axis = [-0.53825, -0.65289, -0.53294]
+    angle = 1.97871
+
+    arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
+    decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
+def test_full_sphere(decomposer_class: Callable[..., Decomposer]) -> None:
```
rturrado, line 33:
```suggestion
def test_all_octants_of_bloch_sphere_rotation(decomposer_class: Callable[..., Decomposer]) -> None:
```

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,44 @@
...
+    decomposer = decomposer_class()
+    axis = [-0.53825, -0.65289, -0.53294]
+    angle = 1.97871
+
+    arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0)
+    decomposed_arbitrary_operation = decomposer.decompose(arbitrary_operation)
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
```
rturrado, line 32: If `test_arbitrary` is no longer needed, I wouldn't use the `DECOMPOSER` constant anymore.

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,44 @@
...
+    YZYDecomposer,
+    ZXZDecomposer,
+    ZYZDecomposer,
+)
+from opensquirrel.passes.decomposer.general_decomposer import check_gate_replacement
+
+DECOMPOSER = [XYXDecomposer, XZXDecomposer, YXYDecomposer, YZYDecomposer, ZXZDecomposer, ZYZDecomposer]
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
```
rturrado, line 21: Do we still need the `test_arbitrary` if we have the `test_full_sphere`?

rturrado (changes requested):

test/decomposer/test_aba_decomposer.py
juanboschero: test_arbitrary is gone but I kept test_specific because it is very precise floats and also the test required for one of our stakeholders.

test/decomposer/test_aba_decomposer.py
juanboschero: I will keep test_specific which does need it

test/decomposer/test_mckay_decomposer.py
juanboschero: I will delete

test/decomposer/test_aba_decomposer.py
```diff
@@ -39,6 +39,6 @@ def test_full_sphere(decomposer_class: Callable[..., Decomposer]) -> None:
 
     for angle in angles:
         for axis in axis_list:
-            arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0.0)
+            arbitrary_operation = BlochSphereRotation(qubit=0, axis=axis, angle=angle, phase=0.1)
```
rturrado, line 41: Why is this change of phase from 0.0 to 0.1 needed?

opensquirrel/passes/decomposer/general_decomposer.py
```diff
@@ -18,8 +20,10 @@ def decompose(self, gate: Gate) -> list[Gate]:
 def check_gate_replacement(gate: Gate, replacement_gates: Iterable[Gate]) -> None:
     gate_qubit_indices = [q.index for q in gate.get_qubit_operands()]
     replacement_gates_qubit_indices = set()
+    replaced_matrix = get_circuit_matrix(get_reindexed_circuit([gate], gate_qubit_indices))
+    norm_matrix = np.abs(replaced_matrix)
 
-    if gate.is_identity():
+    if np.allclose(norm_matrix, np.eye(norm_matrix.shape[0]), atol=ATOL):
         return
```
rturrado, line 27: Shouldn't it be good to keep this check at the very beginning of the function?

opensquirrel/passes/decomposer/general_decomposer.py
```diff
@@ -18,8 +20,10 @@ def decompose(self, gate: Gate) -> list[Gate]:
 def check_gate_replacement(gate: Gate, replacement_gates: Iterable[Gate]) -> None:
     gate_qubit_indices = [q.index for q in gate.get_qubit_operands()]
     replacement_gates_qubit_indices = set()
+    replaced_matrix = get_circuit_matrix(get_reindexed_circuit([gate], gate_qubit_indices))
+    norm_matrix = np.abs(replaced_matrix)
 
-    if gate.is_identity():
+    if np.allclose(norm_matrix, np.eye(norm_matrix.shape[0]), atol=ATOL):
```
rturrado: Can we make this two lines into a function? Beware name `is_normalized` may be incorrect. E.g.:

```
def is_normalized(matrix: Matrix):
    norm_matrix = np.abs(matrix)
    return np.allclose(norm_matrix, np.eye(norm_matrix.shape[0]), atol=ATOL):

if matrix_is_normalized(replaced_matrix):
    return
```

rturrado (changes requested):

test/decomposer/test_aba_decomposer.py
juanboschero: Yes, it tests out the new code fragment in check_gate_replacement

opensquirrel/utils/identity_filter.py
```diff
@@ -3,9 +3,19 @@
...
     from opensquirrel.ir import Gate
 
 
 def filter_out_identities(gates: Iterable[Gate]) -> list[Gate]:
     return [gate for gate in gates if not gate.is_identity()]
+
+
+def is_matrix_identity(matrix: ArrayLike) -> bool:
+    norm_matrix = np.abs(matrix)
+    return np.allclose(norm_matrix, np.eye(norm_matrix.shape[0]), atol=ATOL)
```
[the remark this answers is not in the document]
juanboschero: Yes we need to do the abs because the matrix can be a complex number. This is what was going wrong before with the phase. If you change the phase, you introduce a global phase of e^(phi * i). When the `is_identity` function would be called, it would return `False`. This extra step makes sure that the global phase is accounted for, whether it be negative or complex.

test/decomposer/test_aba_decomposer.py
juanboschero: A change from 0.0 to 0.1 tests out whether the `check_gate_replacement` can account for global phases, more specifically, if it is capable of identifying whether a matrix is an identity gate with a multiplied complex or negative factor. Before, this was not being picked up, thus [[i,0],[0,i]] was returning false to being an identity gate.

opensquirrel/utils/identity_filter.py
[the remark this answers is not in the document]
rturrado: OK, can we add a parametrized test for `is_identity_matrix` with a good deal of cases?

My point is that `is_identity_matrix([[-1, 0], [0, 1]])` is returning true but it shouldn't should it?

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,43 @@
...
+    aba.XYXDecomposer,
+    aba.XZXDecomposer,
+    aba.YXYDecomposer,
+    aba.YZYDecomposer,
+    aba.ZXZDecomposer,
+    aba.ZYZDecomposer,
+]
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
```
rturrado, line 20:
```suggestion
@pytest.mark.parametrize("aba_decomposer", DECOMPOSER)
```

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,43 @@
+from typing import Callable
+
+import numpy as np
+import pytest
+
+from opensquirrel.ir import BlochSphereRotation
+from opensquirrel.passes.decomposer import Decomposer, aba_decomposer as aba
+from opensquirrel.passes.decomposer.general_decomposer import check_gate_replacement
+
+DECOMPOSER = [
```
rturrado, line 10:
```suggestion
ABA_DECOMPOSER_LIST = [
```

test/decomposer/test_aba_decomposer.py
rturrado: Thanks  for the explanation. But I think that either we are mixing tests here, or we are not being exhaustive enough.

Can we:
1. Either have this `test_all_octants_of_bloch_sphere_rotation` focus on axis/angles combos, and then have another `test_all_phases_of_bloch_sphere_rotation` focus on phases?
2. Or, maybe better, add `phase` to the combinations here. I.e., we are testing a combination of axes and angles. Why not multiplying that combination to different phases?

test/decomposer/test_aba_decomposer.py
```diff
@@ -0,0 +1,43 @@
...
+    check_gate_replacement(arbitrary_operation, decomposed_arbitrary_operation)
+
+
+@pytest.mark.parametrize("decomposer_class", DECOMPOSER)
+def test_all_octants_of_bloch_sphere_rotation(decomposer_class: Callable[..., Decomposer]) -> None:
+    decomposer = decomposer_class()
+    steps = 6
+    coordinates = np.linspace(-1, 1, num=steps)
+    angles = np.linspace(-2 * np.pi, 2 * np.pi, num=steps)
+    axis_list = [[i, j, z] for i in coordinates for j in coordinates for z in coordinates]
```
rturrado, line 37:
```suggestion
    steps = 6
    coordinates = np.linspace(-1, 1, num=steps)
    angles = np.linspace(-2 * np.pi, 2 * np.pi, num=steps)
    axes= [[i, j, z] for i in coordinates for j in coordinates for z in coordinates]
```

opensquirrel/passes/decomposer/aba_decomposer.py
```diff
@@ -59,53 +59,40 @@ def get_decomposition_angles(self, alpha: float, axis: AxisLike) -> tuple[float,
...
-            else:
-                acos_argument = float(b_axis_value) * math.sin(alpha / 2) / math.sin(theta2 / 2)
-
-                # This fixes float approximations like 1.0000000000002, which acos does not like.
-                acos_argument = max(min(acos_argument, 1.0), -1.0)
-                m = 2 * math.acos(acos_argument)
-                if math.pi - abs(m) > ATOL:
-                    m_sign = 2 * math.atan2(c_axis_value, a_axis_value)
-                    m = math.copysign(m, m_sign)
+            m = 2 * math.acos(acos_argument)
```
rturrado, line 79: This code is being repeated in two different places. Can we write and use an inner function like?

```
        def _acos(value: float) -> float:
            # This fixes float approximations like 1.0000000000002, which acos does not like.
            value = max(min(value, 1.0), -1.0)
            return math.acos(value)
```

opensquirrel/passes/decomposer/aba_decomposer.py
```diff
@@ -59,53 +59,40 @@ def get_decomposition_angles(self, alpha: float, axis: AxisLike) -> tuple[float,
...
         if is_sin_m_negative:
             m = m * -1
 
         theta1 = (p + m) / 2
         theta3 = p - theta1
 
+        if all(component < 0 for component in axis) or sign:
+            theta1, theta3 = theta3, theta1
+
         return theta1, theta2, theta3
```
rturrado, line 96: Just a general comment on this function:

- The code doesn't seem to follow any logic. It is difficult to understand and to maintain.
- It has changed many times. Which is probably a symptom that is not well tested. So new errors lead to new code.
- Every time it has changed, it has been greatly rewritten, most of the times becoming more complex.

Do you think we could get an understandable version of this code?

rturrado (changes requested):

opensquirrel/passes/decomposer/aba_decomposer.py
rturrado: I've created PR #392 to try to add some logic to the `get_decomposition_angles`  function (until a point, because I see there is a lot of maths behind it).

opensquirrel/common.py
```diff
@@ -49,3 +49,14 @@ def are_matrices_equivalent_up_to_global_phase(
     phase_difference = matrix_a[first_non_zero] / matrix_b[first_non_zero]
 
     return np.allclose(matrix_a, phase_difference * matrix_b)
+
+
+def is_identity_matrix(matrix: NDArray[np.complex128]) -> bool:
```
rturrado, line 54:
```suggestion
def is_identity_matrix_up_to_a_global_phase(matrix: NDArray[np.complex128]) -> bool:
```

opensquirrel/common.py
```diff
@@ -49,3 +49,14 @@ def are_matrices_equivalent_up_to_global_phase(
...
+    """Checks whether matrix is an identity matrix up to a global phase.
+
+    Args:
+        matrix: matrix to check.
+    Returns:
+        Whether matrix is an identity matrix up to a global phase.
+    """
+    return are_matrices_equivalent_up_to_global_phase(matrix, np.eye(matrix.shape[0]))
```
rturrado, line 62: Beautiful. Can we move this function to common.py?

test/decomposer/test_mckay_decomposer.py
```diff
@@ -69,12 +69,15 @@ def test_hadamard(decomposer: McKayDecomposer) -> None:
 
 def test_all_octants_of_bloch_sphere_rotation(decomposer: McKayDecomposer) -> None:
     steps = 6
+    phase_steps = 3
     coordinates = np.linspace(-1, 1, num=steps)
     angles = np.linspace(-2 * np.pi, 2 * np.pi, num=steps)
+    phases = np.linspace(-np.pi, np.pi, num=phase_steps)
```
rturrado, line 75: Nice!

rturrado (changes requested):

opensquirrel/common.py
juanboschero: It is in common.py already

opensquirrel/common.py
```diff
@@ -28,7 +28,7 @@ def normalize_angle(x: SupportsFloat) -> float:
 
 
 def are_matrices_equivalent_up_to_global_phase(
-    matrix_a: NDArray[np.complex128], matrix_b: NDArray[np.complex128]
+    matrix_a: NDArray[np.complex128], matrix_b: NDArray[np.complex128 | np.float64]
```
rturrado, line 31: I understand we are adding this `np.float64` because `np.eye(matrix.shape[0])` uses `np.float64` by default. But we can also keep this function as it is and use `np.eye(matrix.shape[0], dtype=np.complex128)`. Which makes more sense, because quantum matrices are supposed to have complex elements.

```suggestion
    matrix_a: NDArray[np.complex128], matrix_b: NDArray[np.complex128]
```

opensquirrel/common.py
```diff
@@ -49,3 +49,14 @@ def are_matrices_equivalent_up_to_global_phase(
...
+
+def is_identity_matrix_up_to_a_global_phase(matrix: NDArray[np.complex128]) -> bool:
+    """Checks whether matrix is an identity matrix up to a global phase.
+
+    Args:
+        matrix: matrix to check.
+    Returns:
+        Whether matrix is an identity matrix up to a global phase.
+    """
+    return are_matrices_equivalent_up_to_global_phase(matrix, np.eye(matrix.shape[0]))
```
rturrado, line 62:
```suggestion
    return are_matrices_equivalent_up_to_global_phase(matrix, np.eye(matrix.shape[0], dtype=np.complex128))
```

rturrado (changes requested):
Minor final comments.

rturrado (approved):
Beautiful. Thanks for your patience!

PR #380, opened 2024-11-19
[CQT-146] Redesign structure of Parser and CircuitBuilder wrt the InstructionLibrary

rturrado:
Reimplement InstructionLibrary as a singleton:
- Offering a get_gate_set and a get_non_unitary_set API.
- On top of the previously existing get_gate_f and get_non_unitary_f.

Parser and CircuitBuilder just use the InstructionLibrary singleton.

Add Singleton at utils/singleton.py.

Pass tests and linters.

Extra (not belonging to this PR):
- Small tweaks in utils/check_mapper.py.
- Update utils/__init__.py.
- Remove utils/checks folder.


opensquirrel/utils/check_mapper.py
```diff
@@ -10,10 +10,22 @@
...
+    ir_copy = deepcopy(circuit.ir)
+    circuit.map(mapper)
+    assert circuit.ir == ir_copy, "A Mapper pass should not change the IR"
+
+
 def check_mapper(mapper: Mapper) -> None:
     """Check if the `mapper` complies with the OpenSquirrel requirements.
 
-    If an implementation of ``Mapper`` passes these checks it should be compatible with the ``Circuit.map`` method.
+    If ``Mapper`` implementation passes these checks, it should be compatible with the ``Circuit.map`` method.
```
rturrado, line 28:
```suggestion
    If a ``Mapper`` implementation passes these checks, it should be compatible with the ``Circuit.map`` method.
```

opensquirrel/instruction_library.py
```diff
@@ -3,18 +3,23 @@
...
+class InstructionLibrary(metaclass=Singleton):
+    def __init__(self) -> None:
+        self.gate_set = default_gate_set
+        self.non_unitary_set = default_non_unitary_set
+
+    def get_gate_set(self) -> Mapping[str, Callable[..., Gate]]:
+        return self.gate_set
+
+    def get_non_unitary_set(self) -> Mapping[str, Callable[..., NonUnitary]]:
+        return self.non_unitary_set
```
elenbaasc, line 22: Could these be set up as properties?
```suggestion
    def __init__(self) -> None:
        self._gate_set = default_gate_set
        self._non_unitary_set = default_non_unitary_set
        
    @property    
    def gate_set(self) -> Mapping[str, Callable[..., Gate]]:
        return self._gate_set
    
    @gate_set.setter
    def gate_set(self, value: Mapping[str, Callable[..., Gate]]) -> None:
        self._gate_set = value

    @property
    def non_unitary_set(self) -> Mapping[str, Callable[..., NonUnitary]]:
        return self._non_unitary_set
    
    @gate_set.setter
    def non_unitary_set(self, value: Mapping[str, Callable[..., NonUnitary]]) -> None:
        self._non_unitary_set= value
```

opensquirrel/utils/singleton.py
```diff
@@ -0,0 +1,14 @@
...
+_T = t.TypeVar("_T")
+
+
+class Singleton(type, t.Generic[_T]):
+    _instances: t.ClassVar[dict[type, t.Any]] = {}
+
+    def __call__(cls, *args: t.Any, **kwargs: t.Any) -> _T:
+        if cls not in cls._instances:
+            cls._instances[cls] = super().__call__(*args, **kwargs)
+        return t.cast(_T, cls._instances[cls])
```
elenbaasc, line 14: Wow, what is going on here? Looks impressive 😅 Can you explain a bit what is going on?

Is this something that can imported?

elenbaasc (commented):
Looks very nice!

I'm not sure how the Singleton metaclass works, but I'm keen to learn!

opensquirrel/utils/singleton.py
rturrado: To be honest:

- I just copy-pasted a solution from StackOverflow. There are many implementations of a Singleton in Python, but it seems the one using metaclasses is the one with more pros and less cons.
- From there, again using StackOverflow, I found a new version with type annotations that would pass `mypy --strict`.
- Finally, I used ChatGPT to make that code pass `ruff check --fix`
important thing is that:
- We now have a Singleton class in `utils/singleton.py` that we can reuse for other occassions.

And yes, definitely, we can reuse this Singleton class from `utils/singleton.py` if we need more classes in the codebase to follow the Singleton pattern.

In general, I would read that code as:
- I'm a Singleton class, and you can inherit from me if you want to implement a class following the Singleton pattern.
- What does that mean? That there is going to be only one instance of that subclass during runtime.
- How will I do it?
  - First, I will keep a dictionary where the key is the subclass type (e.g., InstructionLibrary) and the value is an instance of that subclass type.
  - Then, I will implement the __call__ method, so you will be able to use your subclass as a functor, e.g., InstructionLibrary(), to get the single instance of your subclass.
  - Finally, how will I implement that __call__ method? If I have previously created an instance of your subclass, I will just return it, otherwise I will create it, save it in the internal dictionary, and return it.

opensquirrel/circuit_builder.py
```diff
@@ -43,27 +42,21 @@ class CircuitBuilder(InstructionLibrary):
...
         self.register_manager = RegisterManager(QubitRegister(qubit_register_size), BitRegister(bit_register_size))
         self.ir = IR()
 
     def __getattr__(self, attr: Any) -> Callable[..., Self]:
         return partial(self._add_instruction, attr)
 
     def _add_instruction(self, attr: str, *args: Any) -> Self:
-        if attr in self.gate_set:
-            generator_f_gate = self.get_gate_f(attr)
+        instruction_library = InstructionLibrary()
```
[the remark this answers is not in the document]
rturrado: That's right. `InstructionLibrary()` ends up executing `Singleton.__call__` which is where you say:

- Do I have an instance of InstructionLibrary already? Otherwise I create that single instance first.
- Then I return the single instance.

opensquirrel/instruction_library.py
rturrado: Hm... I don't really want to offer a setter.

What would be the benefit then to use a property? Just that you would use `InstructionLibrary().gate_set` instead of `InstructionLibrary.get_gate_set()`?

opensquirrel/circuit_builder.py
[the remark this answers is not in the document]
rturrado: In C++ I always implement Singleton's using a `get_instance` static method. So you would do something like `InstructionLibrary.get_instance()` instead of `InstructionLibrary()`. If you also think that is clearer, we can change the code to do that. `get_instance` implementation would be the same than that of `__call__`.

opensquirrel/instruction_library.py
elenbaasc: That would indeed be the benefit. But it is also generally the way to implement getter functionality in Python.

You don't need to offer a setter that's fine.

opensquirrel/circuit_builder.py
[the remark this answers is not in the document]
elenbaasc: A `module` can generally be considered a singleton in Python, so we could also have the instruction library as a module and not implement it as a class.

E.g. `instruction_library.py`:
```
from __future__ import annotations

from collections.abc import Callable, Mapping
from typing import TYPE_CHECKING

from opensquirrel.default_instructions import default_gate_set, default_non_unitary_set

if TYPE_CHECKING:
    from opensquirrel.ir import Gate, NonUnitary


gate_set = default_gate_set
non_unitary_set = default_non_unitary_set

def get_gate_f(gate_name: str) -> Callable[..., Gate]:
    if gate_name not in gate_set:
        msg = f"unknown gate '{gate_name}'"
        raise ValueError(msg)
    return gate_set[gate_name]

def get_non_unitary_f(non_unitary_name: str) -> Callable[..., NonUnitary]:
    if non_unitary_name not in non_unitary_set:
        msg = f"unknown non-unitary instruction '{non_unitary_name}'"
        raise ValueError(msg)
    return non_unitary_set[non_unitary_name]
```
This can then be imported into other modules:

```
import instruction_library

...
gate_set = instruction_library.gate_set
...
generator_f_gate = instruction_library.get_gate_f(attr)
...

File: opensquirrel/utils/singleton.py (hunk 3 above)
Comment (elenbaasc) [reply opensquirrel/utils/singleton.py:14]: Thanks for the detailed explanation!

I would propose to import the classes from typing (consistent with how we do this in the rest of the code):

```suggestion
from typing import TypeVar, Generic, ClassVar, Any, cast

_T = TypeVar("_T")

class Singleton(type, Generic[_T]):
    _instances: ClassVar[dict[type, Any]] = {}

    def __call__(cls, *args: Any, **kwargs: Any) -> _T:
        if cls not in cls._instances:
            cls._instances[cls] = super().__call__(*args, **kwargs)
        return cast(_T, cls._instances[cls])
```

rturrado: Done!

opensquirrel/instruction_library.py
rturrado: Done. Without settters.

opensquirrel/circuit_builder.py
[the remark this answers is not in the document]
rturrado: As a matter of fact, there are many ways to implement this controversial design pattern in Python. I just read in a couple of places that the metaclass approach was the preferred way (one of those places, [here](https://stackoverflow.com/questions/6760685/what-is-the-best-way-of-implementing-singleton-in-python), where the second most voted answer is to use a module).

I personally see these pros and cons to each solution.

### Metaclass
- Pros:
  - I like the more object-oriented solution of using an InstructionLibrary class (encapsulation, consistency with the rest of the code, where we have Circuit's and CircuitBuilder's, and RegisterManager's, all classes).
  - I like the explicitness of declaring `class InstructionLibrary(metaclass=Singleton)`.
- Cons:
  - `Singleton` class implementation is complex, mainly due to type annotation.
  - I would very much prefer to be able to get an instance of a singleton class via `get_instance()` than via `__call__`.

### Module
- Pros:
  - Really simple. You get rid of the `utils/singleton.py`, the `Singleton` class... everything.
- Cons:
  - I find it confusing to read `instruction_library.gate_set` and think `instruction_library` is an object, while it really is a module.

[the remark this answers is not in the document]
rturrado: Done. You beat me. I cannot say no to removing code, especially complex code. 😀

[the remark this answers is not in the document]
elenbaasc: Thanks for looking into this in detail! Great to know about the alternatives 😃

elenbaasc (approved):

PR #391, opened 2024-11-27
Rewrite get_decomposition_angles

rturrado:
Rewrite get_decomposition_angles:
- Move complex code to functions trying to:
  - make the main function more readable, but also
  - remove duplicated code (e.g., acos).
- Add some more comments (e.g., when setting the A, B, and C axes values).

Add utils/math.py with acos and are_axes_consecutive.

Use digit separators for ATOL.


rturrado:
Wrong target branch (`develop`).

PR #394, opened 2024-11-29
[CQT-278] Fix `compose_bloch_sphere_rotations`

rturrado:
`compose_bloch_sphere_rotations` now:
- Keeps the generator and the arguments when combining 2 Bloch sphere rotations with the same generator (same named gate).
- Keeps the arguments when combining 2 Bloch sphere rotations with different generator.
- Tries naming an anonymous BSR before returning it.

This PR includes additional, somewhat unrelated changes:
- Remove BlochSphereRotation.identity.
- Change implementation of `Gate.is_anonymous`: check for a None `generator` instead of a None `arguments`.
- Reimplement `test_all_octants_of_bloch_sphere_rotations`.
- Add `ABADecomposer._are_b_and_c_axes_in_negative_octant`.

---

TODO: this code in `compose_bloch_sphere_rotations` is really awful:

```
if bsr_a.arguments is not None and len(bsr_a.arguments) == 2 and isinstance(bsr_a.arguments[1], Float):
    arguments += (Float(combined_angle),)
```

- It is assuming that, if a BSR has a second argument, that argument represents an angle.
- I don't know if we can do much better with the current generators implementation.
 - I will open a ticket to try to replace the generators implementation, which relies a lot on introspection with the single benefit (I think) of allowing the circuit builder to have a simple syntax.


opensquirrel/passes/merger/general_merger.py
```diff
@@ -1,23 +1,25 @@
...
+from opensquirrel.default_instructions import default_bloch_sphere_rotation_set
+from opensquirrel.ir import IR, Barrier, BlochSphereRotation, Float, Instruction, Statement
 from opensquirrel.utils import flatten_list
 
 
 def compose_bloch_sphere_rotations(a: BlochSphereRotation, b: BlochSphereRotation) -> BlochSphereRotation:
     """Computes the Bloch sphere rotation resulting from the composition of two Bloch sphere rotations.
     The first rotation is applied and then the second.
-    The resulting gate is anonymous except if `a` is the identity and `b` is not anonymous, or vice versa.
+    The resulting Bloch sphere rotation is always tried to be named, in case it were anonymous.
```
elenbaasc, line 20:
```suggestion
    In case the resulting Bloch sphere rotation is anonymous, an attempt is always made to name it.
```

juanboschero, line 20:
```suggestion
    The resulting Bloch sphere rotation is always assigned a name if it would otherwise remain unnamed.
```

opensquirrel/passes/merger/general_merger.py
```diff
@@ -1,23 +1,25 @@
...
 
-    Uses Rodrigues' rotation formula, see for instance https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula.
+    Uses Rodrigues' rotation formula (see https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula).
     """
     if a.qubit != b.qubit:
         msg = "cannot merge two BlochSphereRotation's on different qubits"
```
juanboschero, line 25:
```suggestion
        msg = "cannot merge two BlochSphereRotations on different qubits"
```

opensquirrel/passes/merger/general_merger.py
```diff
@@ -71,14 +114,17 @@ def try_name_anonymous_bloch(bsr: BlochSphereRotation) -> BlochSphereRotation:
...
-            return gate
+    if not bsr.is_anonymous:
+        return bsr
+    for default_bsr_callable in default_bloch_sphere_rotation_set.values():
+        if can_invoke_bsr_callable(default_bsr_callable, bsr):
+            default_bsr = default_bsr_callable(*bsr.get_qubit_operands())
+            if (
+                np.allclose(default_bsr.axis, bsr.axis, atol=ATOL)
+                and np.allclose(default_bsr.angle, bsr.angle, atol=ATOL)
+                and np.allclose(default_bsr.phase, bsr.phase, atol=ATOL)
```
elenbaasc, line 125: These are just a single value, so no need to use `np.allclose()`
```suggestion
                and abs(default_bsr.angle - bsr.angle) < ATOL
                and abs(default_bsr.phase - bsr.phase) < ATOL
```

opensquirrel/utils/check_mapper.py
```diff
@@ -39,7 +40,7 @@ def check_mapper(mapper: Mapper) -> None:
 
     ir = IR()
     ir.add_gate(BlochSphereRotation(42, (1, 0, 0), 1, 2))
-    ir.add_gate(ControlledGate(42, BlochSphereRotation.identity(100)))
+    ir.add_gate(ControlledGate(42, I(100)))
```
juanboschero, line 43: This function looks like a test more than a utility function, can you explain why these numbers are important?

juanboschero (changes requested):
Looking good! Minor comments...

elenbaasc (commented):
Nice work!

opensquirrel/utils/check_mapper.py
rturrado: I cannot because I didn't write the function. I have only changed `BlochSphereRotation.identity` for `I` because I have removed `BlochSphereRotation.identity`.

opensquirrel/passes/merger/general_merger.py
rturrado: It's the Dutch `'s`, like in `foto's` 😀. I don't like writing `BlochSphereRotations` because that class doesn't exist. But I can write `two Bloch sphere rotations`.

opensquirrel/passes/merger/general_merger.py
rturrado: I think I won't use the passive form: `If the final Bloch sphere rotation is anonymous, we will try to match it to a default gate`.

opensquirrel/passes/decomposer/aba_decomposer.py
```diff
@@ -54,6 +55,23 @@ def _set_a_b_c_axes_values(self, axis: AxisLike) -> tuple[Any, Any, Any]:
         _axis = Axis(axis)
         return _axis[self.index_a], _axis[self.index_b], _axis[self._find_unused_index()]
 
+    @staticmethod
+    def _are_b_and_c_axes_in_negative_octant(b_axis_value: float, c_axis_value: float) -> bool:
+        """Given an ABC axis system, and the values for axes B and C.
+        Checks if the values for the B and C axes fall in one of the two negative octants:
+        - one where the value of A is positive, and the values of B and C are negative (or one of them zero).
+        - one where the value of A is negative, and the values of B and C are negative (or one of them zero).
```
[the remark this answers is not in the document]
rturrado: Yes, that text is quite obscure. I just wanted to point out that the difference between the two octants is that A is positive in one and negative in the other. I'll rewrite it.

opensquirrel/passes/merger/general_merger.py
```diff
@@ -1,23 +1,25 @@
...
+from opensquirrel import I
 from opensquirrel.common import ATOL
-from opensquirrel.default_instructions import default_bloch_sphere_rotation_without_params_set
-from opensquirrel.ir import IR, Barrier, BlochSphereRotation, Instruction, Statement
+from opensquirrel.default_instructions import default_bloch_sphere_rotation_set
+from opensquirrel.ir import IR, Barrier, BlochSphereRotation, Float, Instruction, Statement
 from opensquirrel.utils import flatten_list
 
 
 def compose_bloch_sphere_rotations(a: BlochSphereRotation, b: BlochSphereRotation) -> BlochSphereRotation:
```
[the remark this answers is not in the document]
rturrado: Fair enough.

opensquirrel/passes/merger/general_merger.py
```diff
@@ -48,16 +50,57 @@ def compose_bloch_sphere_rotations(a: BlochSphereRotation, b: BlochSphereRotatio
 
     combined_phase = np.round(a.phase + b.phase, order_of_magnitude)
 
-    generator = b.generator if a.is_identity() else a.generator if b.is_identity() else None
-    arguments = b.arguments if a.is_identity() else a.arguments if b.is_identity() else None
+    if a.is_identity():
```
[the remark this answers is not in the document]
rturrado: I didn't change this code in this PR, did I?

What do you mean?
- To have another PR implementing correctly `is_identity`?
- To use `is_identity_up_to_a_global_pahse` in line 53 above?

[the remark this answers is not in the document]
elenbaasc: No you did not change this indeed. I just came across it and thought those conditions might be a bit too stringent.

I guess this method is used as a check, to see if a BSR is basically an identity, and therefore can be ignored. Since we, at the moment, do not take the global phase into account in our decompositions, I don't think that our check for whether something is an identity should than be exactly equal to identity (i.e. not differing by a global phase). 

So as part of this PR we could change it to:
```
 def is_identity(self) -> bool:
    # Checks if the Bloch sphere rotation is equal to identity (up to a global phase).
    return abs(self.angle) < ATOL   
```

But we could also do it in another PR (up to you).

opensquirrel/passes/merger/general_merger.py
juanboschero: Maybe the latter is better here hahahahah

opensquirrel/utils/check_mapper.py
juanboschero: Can you see if it necessary to have in the function, if not, do you think it's best to move to a test?

test/merger/test_general_merger.py
```diff
@@ -23,6 +23,33 @@ def test_compose_bloch_sphere_rotations_different_axis() -> None:
...
+        (Rx(0, theta=math.pi), Ry(0, theta=math.pi / 2), BlochSphereRotation(0, axis=(1, 0, 1), angle=math.pi)),
+    ],
+    ids=[
+        "[bsr_a == I]",
+        "[bsr_b == I]",
+        "[bsr_a.generator == bsr_b.generator] X + X == I",
+        "[bsr_a.generator == bsr_b.generator] H + H == I",
+        "[bsr_a.generator == bsr_b.generator] Rx(pi) + Rx(pi) == I",
+        "[bsr_a.generator == bsr_b.generator] Rx(pi/2) + Rx(pi/2) != X",
+        "[bsr_a.generator != bsr_b.generator] Rx(pi) + Ry(pi/2) != H",
```
elenbaasc, line 44: In our Slack conversation we indeed used addition to discuss the various cases, but multiplication is more appropriate. 
```suggestion
        "[bsr_a.generator == bsr_b.generator] X * X == I",
        "[bsr_a.generator == bsr_b.generator] H * H == I",
        "[bsr_a.generator == bsr_b.generator] Rx(pi) * Rx(pi) == I",
        "[bsr_a.generator == bsr_b.generator] Rx(pi/2) * Rx(pi/2) = Rx(pi) ~ X",
        "[bsr_a.generator != bsr_b.generator] Rx(pi) * Ry(pi/2) ~ H",
```
The `~` character is used to denote 'equal up to a global phase'.

elenbaasc (commented):
Thanks for the work, really nice!

Just a minor suggestion in the description of the test cases. So very close to be merged :)

elenbaasc (approved):
All good from my side!

juanboschero (approved):

PR #399, opened 2024-12-05
Change __init__ files

GuyPuts:


opensquirrel/__init__.py
```diff
@@ -1,4 +1,73 @@
...
+    barrier,
+    default_bloch_sphere_rotation_set,
+    default_bloch_sphere_rotation_without_params_set,
+    default_unitary_set,
+    init,
+    measure,
+    mX90,
+    mY90,
+    reset,
+    wait,
```
rturrado, line 33: Do we want to sort these lines?

opensquirrel/__init__.py
```diff
@@ -1,4 +1,73 @@
...
+    "CRk",
+    "SWAP",
+    "measure",
+    "init",
+    "reset",
+    "barrier",
+    "wait",
+    "default_bloch_sphere_rotation_without_params_set",
+    "default_bloch_sphere_rotation_set",
+    "default_unitary_set",
```
rturrado, line 72: Do we want to sort these lines?

opensquirrel/__init__.py
```diff
@@ -1,4 +1,73 @@
...
+    Sdag,
+    T,
+    Tdag,
+    X,
+    Y,
+    Z,
+    barrier,
+    default_bloch_sphere_rotation_set,
+    default_bloch_sphere_rotation_without_params_set,
+    default_unitary_set,
```
rturrado, line 27: I would take these out. They are internal variables, not something a user would need to easily access.

opensquirrel/__init__.py
rturrado, line 72: I think we can take these out.

opensquirrel/passes/merger/general_merger.py
```diff
@@ -6,10 +6,10 @@
 
 import numpy as np
 
+from opensquirrel import default_bloch_sphere_rotation_without_params_set
```
rturrado, line 9: This line should stay as it was.

opensquirrel/utils/__init__.py
```diff
@@ -1,6 +1,18 @@
...
+    "are_axes_consecutive",
+    "filter_out_identities",
+    "flatten_list",
+    "get_matrix",
+    "check_mapper",
+    "expand_ket",
+    "get_reduced_ket",
+    "can1",
+    "get_matrix",
+]
```
rturrado, line 18: Nice! I think, even these identifiers being internal, we want to have them in this init file for our own sake (to make their referencing easier from everywhere).

rturrado: Have we checked also that, everywhere in the code, we just do `from opensquirrel.utils import...` and not `from opensquirrel.utils.blah import...`?

opensquirrel/__init__.py
```diff
@@ -1,4 +1,73 @@
...
+from opensquirrel.writer import writer
 
-__all__ = ["Circuit", "CircuitBuilder"]
+__all__ = [
+    "Circuit",
+    "CircuitBuilder",
+    "writer",
+    "InverseGateModifier",
+    "PowerGateModifier",
+    "ControlGateModifier",
```
rturrado, line 43: I think we can take these out.

rturrado (commented):
Beautiful, I like it!

I've pointed out a few items I would exclude.

opensquirrel/__init__.py
GuyPuts: Regarding this, would it be a good idea to run isort on the whole library? There are more unsorted imports in multiple files; isort would sort and format them alphabetically

opensquirrel/utils/__init__.py
GuyPuts: I've checked it for every import that I added to init. I'll double check just to be sure.

rturrado (approved):
It looks good to me.

But let's anyway wait for Chris to do his review.

Thanks Guy!

elenbaasc (approved):
Looks all good, thanks for the effort!

PR #400, opened 2024-12-06
[CQT-288] Fix bug in ZXZ decomposer

rturrado:
Fix check for needing to swap theta 1 and theta 3.

We were checking (b < 0 and c < 0) instead of (b <= 0 and c <= 0 and not (b ==0 and c == 0)).

Add test_specific_gate to test_zxz_decomposer.py.

Fix test_all_octants_of_bloch_sphere_rotation in test_aba_decomposer.py:
- Change the number of steps for axes from 6 to 5, so  (e.g., 1, 0.5, 0, -0.5, -1).

Update test_cnot_decomposer in test_decompositions.py:
- We had seen this before. We randomly write out Rz(pi) or Rz(-pi).


test/decomposer/test_zxz_decomposer.py
```diff
@@ -46,6 +46,20 @@ def test_zxz_decomposer(decomposer: ZXZDecomposer, gate: Gate, expected_result:
...
+        BlochSphereRotation(0, axis=[0.0, 1.0, 0.0], angle=-math.pi / 2, phase=0.0),
+        BlochSphereRotation(0, axis=[0.0, -1.0, 0.0], angle=math.pi / 2, phase=0.0),
+    ],
+    ids=["Ry(-pi/2)", "BSR_1 of Ry(-pi/2)", "BSR_2 of Ry(-pi/2)"],
+)
+def test_specific_gate(decomposer: ZXZDecomposer, gate: Gate) -> None:
+    decomposed_gate = decomposer.decompose(gate)
+    check_gate_replacement(gate, decomposed_gate)
+
+
```
elenbaasc, line 62: This test was put here to reproduce the bug. Should we make a more general test that checks edge cases? It could be similar to the 'all  octant check' for the McKay decomposer, but that would also add a lot of testing time (the 'all octant check' already takes quite some time now...)

opensquirrel/passes/decomposer/aba_decomposer.py
```diff
@@ -102,7 +102,11 @@ def get_decomposition_angles(self, axis: AxisLike, alpha: float) -> tuple[float,
         theta_3 = p - theta_1
 
         # Check if theta 1 and theta 3 have to be swapped
-        if b_axis_value < 0 and c_axis_value < 0:
+        if (
+            (b_axis_value < 0 or abs(b_axis_value) < ATOL)
+            and (c_axis_value < 0 or abs(c_axis_value) < ATOL)
+            and not (abs(b_axis_value) < ATOL and abs(c_axis_value) < ATOL)
+        ):
```
elenbaasc, line 109: Great catch here!

test/decomposer/test_aba_decomposer.py
```diff
@@ -31,15 +32,17 @@ def test_specific_bloch_rotation(aba_decomposer: Callable[..., Decomposer]) -> N
...
     coordinates = np.linspace(-1, 1, num=steps)
     angles = np.linspace(-2 * np.pi, 2 * np.pi, num=steps)
     phases = np.linspace(-np.pi, np.pi, num=phase_steps)
     axes = [[i, j, z] for i in coordinates for j in coordinates for z in coordinates]
 
-    for angle in angles:
-        for axis in axes:
+    for axis in axes:
+        if all(abs(component) < ATOL for component in axis):  # not a valid axis
+            continue
```
elenbaasc, line 44: Good solution so that we can now include 0 as a value for the axis components (i.e. 5 steps).

elenbaasc (commented):
Thanks for picking this up and resolving the bug!

The windows python builds seem to be failing. Could this be an issue to do with rounding?

test/decomposer/test_zxz_decomposer.py
rturrado: The 'all octants check' is done for the ABA decomposer, Chris, i.e., for all XYX, XZX... ZYZ decomposers.

So we could get rid of this specific one, because it's covered in 'all octants check'.

opensquirrel/passes/decomposer/aba_decomposer.py
rturrado: Lucky!

test/decomposer/test_aba_decomposer.py
rturrado: I think so.

rturrado:
> Thanks for picking this up and resolving the bug!

Thanks! I did because I had seen the code, and it occurred to me that there may be a corner case we were not handling correctly. Which it turned out it was.

> The windows python builds seem to be failing. Could this be an issue to do with rounding?

Yes, kind of, np's definition of atol is stricter than ours (1e-08 instead of 1e-07).

elenbaasc (approved):
Nice!

So why did the windows tests fail?

juanboschero (approved):

PR #405, opened 2024-12-10
Fix: default scheme for light mode

GuyPuts:
Sets the scheme for light mode to default.

### DIFF
--- a/mkdocs-base.yaml
+++ b/mkdocs-base.yaml
@@ -1,0 +1,5 @@
+INHERIT: ./mkdocs-base.yml
+
+extra:
+  version:
+    provider: mike
--- a/mkdocs-base.yaml
+++ b/mkdocs-base.yaml
@@ -1,5 +1,1 @@
 INHERIT: ./mkdocs-base.yml
-
-extra:
-  version:
-    provider: mike
--- a/mkdocs-base.yaml
+++ b/mkdocs-base.yaml
@@ -1,1 +1,0 @@
-INHERIT: ./mkdocs-base.yml
--- a/mkdocs.yaml
+++ b/mkdocs.yaml
@@ -23,8 +23,7 @@ theme:
         name: Switch to light mode
     # Palette toggle for light mode
     - media: "(prefers-color-scheme: light)"
-      primary: white
-      accent: light blue
+      scheme: default
       toggle:
         icon: material/brightness-7
         name: Switch to dark mode


GuyPuts:
As for the third bullet point in the issue, mkdocs.yaml cannot be in docs since mkdocs cannot be run from docs (it must be run from root, where it finds the mkdocs.yaml file and the docs folder). Also, I'm not seeing mkdocs-base.yml, just mkdocs.yaml, is that expected?

rturrado (approved):
Great!

oschusler:
hey @rturrado, thanks for reaching out. My 2 cents on this would be:
1. The distinction between the two files has to do with local and CI/CD setups. For deployment to GitLab/GitHub pages we use Mike. This tool creates versions. However, for local instances, this is highly annoying and you'd much rather have plain mkdocs. With the two file setup you can easily run mkdocs locally and use Mike for pipeline based deployments.
2. I don't necessarily see the mkdocs file as a polution of the root folder. There are more files in the root that address a specific section (`.editorconfig`, `.markdownlint-cli2.jsonc`, `.tmuxinator.yml` or `renovate.json`). Of course, we can move it with a plugin. However, I prefer to have less libraries/plugins/dependencies because there would be more risk of breakages. If we upgrade mkdocs at some point, and this plugin is holding us back, do we really want to run that risk? Basically, my question would be; does the file harm (in the broadest sense of the word) anything being in the root. If not, lets just leave it there.

rturrado:
Great, thanks for the quick answer!

> Basically, my question would be; does the file harm (in the broadest sense of the word) anything being in the root. If not, lets just leave it there.

No, it doesn't. I just thought it was tidier to be able to have those files within `docs`.

So, @GuyPuts , in summary:
1. Let's divide `mkdocs.yaml` into `mkdocs.yaml` and `mkdocs-base.yaml`, as we do for [libQASM](https://github.com/QuTech-Delft/libqasm).
2. Let's leave those files at the root level, above the `docs` folder.

rturrado (changes requested):
@GuyPuts, can we please add the division of the `mkdocs.yaml` into 2 files to this PR?

GuyPuts:
> @GuyPuts, can we please add the division of the `mkdocs.yaml` into 2 files to this PR?

Yes of course. Do we want to include Mike in this as well? The mkdocs.yml in LibQASM only contains the following:
```INHERIT: ./mkdocs-base.yml

extra:
  version:
    provider: mike```

rturrado:
Let's do one thing. Let's keep `mkdocs.yaml` files as they are for the moment. We can handle this split when we add the GitHub workflows for publishing the docs.

rturrado (approved):

GuyPuts:
> Let's do one thing. Let's keep `mkdocs.yaml` file as it is for the moment. We can handle this split when we add the GitHub workflows for publishing the docs.

@rturrado alright! I'll remove the contents of mkdocs-base outside of the INHERIT, and I'll merge :)

rturrado:
> @rturrado alright! I'll remove the contents of mkdocs-base outside of the INHERIT, and I'll merge :)

No, no. Leave `mkdocs.yaml` as it was, without an `mkdocs-base.yaml`.

GuyPuts:
> > @rturrado alright! I'll remove the contents of mkdocs-base outside of the INHERIT, and I'll merge :)
> 
> No, no. Leave `mkdocs.yaml` as it was, without an `mkdocs-base.yaml`.

@rturrado alright!

rturrado (approved):

PR #406, opened 2024-12-11
[CQT-289] Avoid creating an Axis (0, 0, 0)

GuyPuts:
Avoids that an axis of (0, 0, 0) can be created and allows the user to check for this instance before instantiating the class.


opensquirrel/ir.py
```diff
@@ -214,8 +214,29 @@ def __init__(self, *axis: AxisLike) -> None:
...
+
+        Returns:
+            True if the axis is valid, False otherwise.
+        """
+        try:
+            axis_to_check = axis[0] if len(axis) == 1 else cast(AxisLike, axis)
+            axis_array = np.asarray(axis_to_check, dtype=float).flatten()
+            if len(axis_array) != 3:
+                return False
+            return not np.all(axis_array == 0)
```
rturrado, line 236: It's usually preferable to use the positive form, e.g., here:

```suggestion
            return np.any(axis_array != 0)
```

rturrado (commented):
OK, let's have a look at the Axis class globally, and try to refine the logic a bit while not duplicating code.

We have already this function `_parse_and_validate_axislike` that parses and normalizes:
1. I would divide into two functions: one for parsing and the one  for normalizing, e.g., `parse` and `normalize`. Can they both be static methods?
2. I would add the check for !(0, 0, 0) to that `parse` method.

Now, if we go back to the constructor, can we do?
```
axis_to_parse = axis[0] if axis is a list, otherwise axis
self._value = Axis.parse(axis_to_parse)
self._value = Axis.normalize(self._value)
```

We can write 3 parametrized tests for each function: constructor, parse_axis, and normalize_axis.

Let me know how you see it!

opensquirrel/ir.py
GuyPuts: Wouldn't any(axis != 0) mean that none of the values on the Axis can be zero? For example I believe that (1, 0, 1) would be a valid Axis, but it would not pass np.any(axis_array != 0)

GuyPuts:
> OK, let's have a look at the Axis class globally, and try to refine the logic a bit while not duplicating code.
> 
> We have already this function `_parse_and_validate_axislike` that parses and normalizes:
> 
> 1. I would divide it into two functions: one for parsing and the other one for normalizing, e.g., `parse` and `normalize` (this latter already exists). Can they both be static methods?
> 2. I would add the check for `!(0, 0, 0)` to that `parse` method.
> 
> Now, if we go back to the constructor, can we do?
> 
> ```
> axis_to_parse = axis[0] if axis is a list, otherwise axis
> self._value = Axis.parse(axis_to_parse)
> self._value = Axis.normalize(self._value)
> ```
> 
> We can write 3 parametrized tests for each function: constructor, `Axis.parse`, and `Axis.normalize`.
> 
> Let me know how you see it!

@rturrado if we split the functions, parse "should" return the axis object. Shall we return false if the Axis is invalid?

opensquirrel/ir.py
rturrado: No, no, `any(x != 0)` means that "any of the values in Axis should be zero".

For "none of the values in Axis can be zero", i.e., "all of the values in Axis should be non-zero", you would have to write `all(x != 0)`.

GuyPuts:
> > @rturrado if we split the functions, parse "should" return the axis object. Shall we return false if the Axis is invalid?
> 
> We should raise a ValueError.

I agree that usually a ValueError would be best practice. However, the ticket states that we want to create an Axis without a try/catch block:
image

So I believe that in this case, maybe it is better to return False?

rturrado:
No, Guy, let's try to raise the exception. I wrote that ticket. As you can see, I was valuing options, wondering, not stating anything.

GuyPuts:
> No, Guy, let's try to raise the exception. I wrote that ticket. As you can see, I was valuing options, wondering, not stating anything.

Ah alright! I'll raise the exception :)

opensquirrel/ir.py
```diff
@@ -255,10 +257,13 @@ def _parse_and_validate_axislike(cls, axis: AxisLike) -> NDArray[np.float64]:
         if len(axis) != 3:
             msg = f"axis requires an ArrayLike of length 3, but received an ArrayLike of length {len(axis)}"
             raise ValueError(msg)
-        return cls._normalize_axis(axis)
+        if np.any(axis != 0).item():
+            return axis
+        msg = "axis requires at least one element to be non-zero"
+        raise ValueError(msg)
```
rturrado, line 263: This looks a bit ugly. You read it as:

```
if the axis doesn't have exactly 3 components:
    raise an error saying the axis should have exactly 3 components
if any of the components is non-zero:
    return axis
raise an error saying at least one component should be non-zero
```

So, the last error is kind of a method default case, but it's not, really. I would write, instead:

```
if error case 1:
    raise error case 1 message
if error case 2:
    raise error case 2 message
return axis
```

I.e.:

```
        if np.any(axis == 0).item():
            msg = "axis requires at least one element to be non-zero"
            raise ValueError(msg)
        return axis
```

opensquirrel/ir.py
```diff
@@ -214,7 +214,8 @@ def __init__(self, *axis: AxisLike) -> None:
         axis: An ``AxisLike`` to create the axis from.
         """
         axis_to_parse = axis[0] if len(axis) == 1 else cast(AxisLike, axis)
-        self._value = self._parse_and_validate_axislike(axis_to_parse)
+        self._value = self.parse(axis_to_parse)
+        self._value = self.normalize(self._value)
```
rturrado, line 218:
```suggestion
        self._value = self.normalize(self.parse(axis_to_parse))
```

test/test_ir.py
```diff
@@ -93,6 +94,63 @@ def test_eq_true(self, axis: Axis, other: Any) -> None:
...
+        [
+            ([1, 0, 0], np.array([1, 0, 0], dtype=np.float64)),
+            ([0, 0, 0], ValueError),
+            ([1, 2], ValueError),
+            ([1, 2, 3, 4], ValueError),
+            ([0, 1, 0], np.array([0, 1, 0], dtype=np.float64)),
+            (["a", "b", "c"], TypeError),
+        ],
+    )
+    def test_parser(self, axis: AxisLike, expected: Any) -> None:
```
rturrado, line 128:
```suggestion
    def test_parse(self, axis: AxisLike, expected: Any) -> None:
```

test/test_ir.py
```diff
@@ -44,7 +45,7 @@ def test_axis_getter(self, axis: Axis) -> None:
             (Axis(0, 1, 0), [0, 1, 0]),
         ],
     )
-    def test_axis_setter_no_error(self, axis: Axis, new_axis: ArrayLike, expected_axis: ArrayLike) -> None:
+    def test_axis_setter_no_error(self, axis: Axis, new_axis: AxisLike, expected_axis: list[float]) -> None:
```
rturrado, line 48: Have you tried setting a `[0, 0, 0]` value?

rturrado (commented):
Good work, Guy.

I've added a few comments.

opensquirrel/ir.py
```diff
@@ -257,10 +256,10 @@ def parse(cls, axis: AxisLike) -> NDArray[np.float64]:
         if len(axis) != 3:
             msg = f"axis requires an ArrayLike of length 3, but received an ArrayLike of length {len(axis)}"
             raise ValueError(msg)
-        if np.any(axis != 0).item():
-            return axis
-        msg = "axis requires at least one element to be non-zero"
-        raise ValueError(msg)
+        if np.all(axis == 0).item():
```
rturrado, line 259: Why do we need the `.item()` here, @GuyPuts?

```suggestion
        if np.all(axis == 0):
```

rturrado (approved):

PR #407, opened 2024-12-12
[CQT-259] Integrate wait and barrier directives in cQASMv1 exporter

elenbaasc:
- Restore SGMQ notation for barrier links.
- Add tests for barrier links.
- Add CHANGELOG (and entry).


CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+
+## [ 0.2.0 ] - [ xxxx-yy-zz ]
+
+### Added
+- Restoring SGMQ notation for barrier links in cQASMv1 Exporter (with tests).
```
rturrado, line 10:
```suggestion
- Restore SGMQ notation for barrier links in cQASMv1 Exporter (with tests).
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -14,12 +15,13 @@ class _CQASMv1Creator(IRVisitor):
     # Precision used when writing out a float number
     FLOAT_PRECISION = 8
 
-    def __init__(self, register_manager: RegisterManager) -> None:
+    def __init__(self, register_manager: RegisterManager, barrier_links: list[list[Statement]]) -> None:
         self.register_manager = register_manager
         qubit_register_size = self.register_manager.get_qubit_register_size()
```
rturrado, line 20: I'll open a ticket to implement RegisterManager as a singleton.

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -14,12 +15,13 @@ class _CQASMv1Creator(IRVisitor):
...
-        self.cqasmv1_string = "version 1.0\n\n{}\n\n".format(
+        self.cqasm_v1_string = "version 1.0\n\n{}\n\n".format(
```
rturrado, line 21: Should we just use `output`, as in the Writer? Simpler, consistent with the writer.

```suggestion
        self.output = "version 1.0\n\n{}\n\n".format(
```

test/exporter/test_cqasmv1_exporter.py
```diff
@@ -236,3 +236,57 @@ def test_anonymous_gates(gate: Gate) -> None:
...
+            """version 3.0\n\nqubit[5] q\n\nbarrier q[0]\nH q[1]\nbarrier q[1]\nbarrier q[2]\nX q[2]
+barrier q[3]\nbarrier q[4]\nbarrier q[1]\nY q[3]\nbarrier q[0]\n""",
+            """version 1.0\n\nqubits 5\n\nbarrier q[0]\nh q[1]\nbarrier q[1, 2]\nx q[2]\nbarrier q[3, 4, 1]\ny q[3]
+barrier q[0]\n""",
+        ),
+        (
+            """version 3.0; qubit[3] q; barrier q[0]; barrier q[1]; X q[2]; barrier q[1]""",
+            """version 3.0\n\nqubit[3] q\n\nbarrier q[0]\nbarrier q[1]\nX q[2]\nbarrier q[1]\n""",
+            """version 1.0\n\nqubits 3\n\nbarrier q[0, 1]\nx q[2]\nbarrier q[1]\n""",
+        ),
```
rturrado, line 277: We shouldn't need:
- triple quotes for single line texts.
- semicolons, whitespaces, or newlines at the end of an input cQASM program.

Double check if exported programs need the newline at the end.

```suggestion
            """version 3.0; qubit[1] q; barrier q[0]",
            """version 3.0\n\nqubit[1] q\n\nbarrier q[0]\n",
            """version 1.0\n\nqubits 1\n\nbarrier q[0]\n",
        ),
        (
            "version 3.0; qubit[3] q; barrier q[0:2]",
            "version 3.0\n\nqubit[3] q\n\nbarrier q[0]\nbarrier q[1]\nbarrier q[2]\n",
            "version 1.0\n\nqubits 3\n\nbarrier q[0, 1, 2]\n",
        ),
        (
            "version 3.0; qubit[1] q; barrier q[0, 0]",
            "version 3.0\n\nqubit[1] q\n\nbarrier q[0]\nbarrier q[0]\n",
            "version 1.0\n\nqubits 1\n\nbarrier q[0, 0]\n",
        ),
        (
            "version 3.0; qubit[6] q; barrier q[0:2, 5, 3, 4, 1]",
            "version 3.0\n\nqubit[6] q\n\nbarrier q[0]\nbarrier q[1]\nbarrier q[2]\nbarrier q[5]\nbarrier q[3]
barrier q[4]\nbarrier q[1]\n",
            "version 1.0\n\nqubits 6\n\nbarrier q[0, 1, 2, 5, 3, 4, 1]\n",
        ),
        (
            "version 3.0; qubit[5] q; barrier q[0]; H q[1]; barrier q[1:2]; X q[2]; barrier q[3:4, 1]; Y q[3];
            barrier q[0]",
            "version 3.0\n\nqubit[5] q\n\nbarrier q[0]\nH q[1]\nbarrier q[1]\nbarrier q[2]\nX q[2]
barrier q[3]\nbarrier q[4]\nbarrier q[1]\nY q[3]\nbarrier q[0]\n",
            "version 1.0\n\nqubits 5\n\nbarrier q[0]\nh q[1]\nbarrier q[1, 2]\nx q[2]\nbarrier q[3, 4, 1]\ny q[3]
barrier q[0]\n",
        ),
        (
            "version 3.0; qubit[3] q; barrier q[0]; barrier q[1]; X q[2]; barrier q[1]",
            "version 3.0\n\nqubit[3] q\n\nbarrier q[0]\nbarrier q[1]\nX q[2]\nbarrier q[1]\n",
            "version 1.0\n\nqubits 3\n\nbarrier q[0, 1]\nx q[2]\nbarrier q[1]\n",
        ),
```

rturrado: For the future, when indices and SGMQ are taken care of, a couple of cases should be looked into:
1. `q[0, 0]`. Wouldn't that be the same as `q[0]`?
2. `q[3, 2, 1]`. Better written as `q[1, 2, 3]`. We cannot do that at the moment for things like `CNOT q[1, 2] q[4, 0]`, where order is important.

rturrado (commented):
Beautiful work. Just some minor comments.

test/exporter/test_cqasmv1_exporter.py
elenbaasc: I'll add this to your document on SGMQ notation, so that we can take it into account when we deal with that.

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -14,12 +15,11 @@ class _CQASMv1Creator(IRVisitor):
     # Precision used when writing out a float number
     FLOAT_PRECISION = 8
 
-    def __init__(self, register_manager: RegisterManager) -> None:
+    def __init__(self, register_manager: RegisterManager, barrier_links: list[list[Statement]]) -> None:
```
rturrado, line 18: This looks like a code smell. We add a new functionality, and we have to upate the exporter's `__init__` function to cope with it?

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -35,24 +35,27 @@ def visit_float(self, f: SupportsFloat) -> str:
...
 
     def visit_reset(self, reset: Reset) -> None:
         qubit_argument = reset.arguments[0].accept(self)  # type: ignore[index]
-        self.cqasmv1_string += f"prep_z {qubit_argument}\n"
+        self.output += f"prep_z {qubit_argument}\n"
 
     def visit_barrier(self, barrier: Barrier) -> None:
-        qubit_argument = barrier.arguments[0].accept(self)  # type: ignore[index]
-        self.cqasmv1_string += f"barrier {qubit_argument}\n"
+        if self.barrier_links and barrier == self.barrier_links[0][-1]:
```
rturrado, line 49: Code like this should be more readable. For example, express what you want to do with a function:

```
if is_last_barrier_in_the_current_barriers_group(barrier):
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -14,12 +15,11 @@ class _CQASMv1Creator(IRVisitor):
...
 
-    def __init__(self, register_manager: RegisterManager) -> None:
+    def __init__(self, register_manager: RegisterManager, barrier_links: list[list[Statement]]) -> None:
         self.register_manager = register_manager
         qubit_register_size = self.register_manager.get_qubit_register_size()
-        self.cqasmv1_string = "version 1.0\n\n{}\n\n".format(
-            f"qubits {qubit_register_size}" if qubit_register_size > 0 else ""
-        )
+        self.output = "version 1.0\n\n{}\n\n".format(f"qubits {qubit_register_size}" if qubit_register_size > 0 else "")
+        self.barrier_links = barrier_links
```
rturrado, line 22: Where does this 'links' terminology come from? In the code, we talk about "groups of 'linked' barriers". But where do we explain what are these `barrier_links`?

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -62,15 +65,17 @@ def visit_gate(self, gate: Gate) -> None:
...
 
 def export(circuit: Circuit) -> str:
-    cqasmv1_creator = _CQASMv1Creator(circuit.register_manager)
-
+    barrier_links = [
+        list(barrier_link)
+        for barrier, barrier_link in groupby(circuit.ir.statements, lambda stmt: isinstance(stmt, Barrier))
+        if barrier
+    ]
+    cqasmv1_creator = _CQASMv1Creator(circuit.register_manager, barrier_links)
```
rturrado, line 77: Why don't we?:
- Keep the visitor simple. For example, when a barrier comes, it will just print it.
- So we do not pass any barrier links to the visitor.
- We cannot just copy the IR and transform it to convert each group of 'linked' barriers into a list, because the visitor interface doesn't understand about nodes that are list of barriers.
- What we can do is a simple post processing (yes, maybe slower, because it's an extra pass, but clearer) which goes over the final program and transforms something like `barrier q[i]\nbarrier q[j]` into `barrier q[i, j]`:

```
def dump_barrier_group(indices: list[int]) -> str:
    return "barrier q[{}]".format(", ".join(indices))

def merge_barrier_groups(output: str) -> str:
    new_output = ""
    barrier_group_indices = []
    for line in output.split('\n'):
        if not line.begin("barrier"):
            new_output += dump_barrier_group(barrier_group_indices)
            new_output += line
            barrier_group_indices = []
        else:
            barrier_group_indices.extend(get_barrier_indices(line))
    if len(barrier_group_indices) != 0:
        new_output += dump_barrier_group(current_barrier_group_indices)
```

Then:

```
    output = cqasmv1_creator.output.rstrip() + "\n"
    return merge_barrier_groups(output)
```

rturrado (changes requested):
I have suggested merging the barrier groups as a post processing, in order to keep the visitor as clean as possible.

test/exporter/test_cqasmv1_exporter.py
```diff
@@ -236,3 +236,57 @@ def test_anonymous_gates(gate: Gate) -> None:
     with pytest.raises(UnsupportedGateError, match="not supported"):  # noqa: PT012
         qc = builder.to_circuit()
         qc.export(fmt=ExportFormat.CQASM_V1)
+
+
+@pytest.mark.parametrize(
+    ("v3_input", "v3_output", "v1_output"),
```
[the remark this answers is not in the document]
elenbaasc: You're right, this was just a check for my own sanity. They can be removed.

opensquirrel/passes/exporter/cqasmv1_exporter.py
elenbaasc: Good point! We should indeed be consistent in our terms. They are indeed groups of linked barriers.

opensquirrel/passes/exporter/cqasmv1_exporter.py
elenbaasc: I didn't want to do it like this, because then the only criteria were using to determine whether something is a barrier is if the string is equal to "barrier". Doesn't _feel_ robust (but I must admit, it is more just a _feeling_). I do like to keep the visitor class clean though, so I'd be happy to implement it like this. What do you think?

opensquirrel/passes/exporter/cqasmv1_exporter.py
elenbaasc: "code smell", I learned a valuable new term today :)

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -62,9 +60,34 @@ def visit_gate(self, gate: Gate) -> None:
...
         qubit_args = (arg.accept(self) for arg in gate.arguments if isinstance(arg, Qubit))  # type: ignore[union-attr]
-        self.cqasmv1_string += "{} {}{}\n".format(
-            gate_name, ", ".join(qubit_args), ", " + ", ".join(params) if params else ""
-        )
+        self.output += "{} {}{}\n".format(gate_name, ", ".join(qubit_args), ", " + ", ".join(params) if params else "")
+
+
+def post_process(output: str) -> str:
+    post_output = _merge_barrier_groups(output)
+    return post_output.rstrip() + "\n"
```
rturrado, line 68: I don't think we need to do the `rstrip() + "\n" part again. We just have to take care not to remove that final newline.

```suggestion
    return _merge_barrier_groups(output)
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -62,9 +60,34 @@ def visit_gate(self, gate: Gate) -> None:
...
+            barrier_group_indices.append(_get_barrier_index(line))
+    return post_output
+
+
+def _dump_barrier_group(indices: list[int]) -> str:
+    return "barrier q[{}]\n".format(", ".join(map(str, indices)) if len(indices) != 0 else "")
+
+
+def _get_barrier_index(line: str) -> int:
+    return int("".join(s for s in line if s.isdigit()))
```
rturrado, line 90: Something like?

```suggestion
import re

def _get_barrier_index(line: str) -> int:
    return int(re.find('\d+', line))
```

rturrado (commented):
I've added a couple of comments.

CHANGELOG.md
```diff
@@ -1,10 +1,18 @@
...
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+### Types of changes:
+* **Added** for new features.
+* **Changed** for changes in existing functionality.
+* **Deprecated** for soon-to-be removed features.
+* **Removed** for now removed features.
+* **Fixed** for any bug fixes.
+* **Security** in case of vulnerabilities.
```
rturrado, line 12: I wouldn't use `Deprecated` and `Security`. If we ever have them, we can add them to the list. For the time being, I would keep things simple.

```suggestion
* **Added** for new features.
* **Changed** for changes in existing functionality.
* **Removed** for now removed features.
* **Fixed** for any bug fixes.
```

CHANGELOG.md
```diff
@@ -0,0 +1,16 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+This project adheres to [Semantic Versioning](http://semver.org/).
+
+### Types of changes:
+* **Added** for new features.
+* **Changed** for changes in existing functionality.
+* **Removed** for now removed features.
+* **Fixed** for any bug fixes.
```
rturrado, line 10: I would keep this order. More than for being an alphabetical order, it also follows some logic "addition, modification, removal", Fixed being a type of Modification so to speak.

```suggestion
* **Added** for new features.
* **Changed** for changes in existing functionality.
* **Fixed** for any bug fixes.
* **Removed** for now removed features.
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -62,9 +61,33 @@ def visit_gate(self, gate: Gate) -> None:
...
+            barrier_group_indices.append(_get_barrier_index(line))
+    return post_output
+
+
+def _dump_barrier_group(indices: list[int]) -> str:
+    return "barrier q[{}]\n".format(", ".join(map(str, indices)) if len(indices) != 0 else "")
+
+
+def _get_barrier_index(line: str) -> int:
+    return int(re.findall("\d+", line)[0])
```
rturrado, line 90: Can we?
1. Use `search` instead of `findall` (since we know there will only be one index occurrence).
2. Add some error checks (since we don't know if there will have been some problems when generating the code).

```suggestion
    barrier_index_match = re.search("\d+", line)
    if not barrier_index_match:
        raise CqasmV1ExporterParseError("expecting a barrier index but found none")       
    return int(barrier_index_match.group(0))
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -62,9 +61,33 @@ def visit_gate(self, gate: Gate) -> None:
...
+    barrier_group_indices: list[int] = []
+    for line in output.split("\n"):
+        if not line.startswith("barrier"):
+            if barrier_group_indices:
+                post_output += _dump_barrier_group(barrier_group_indices)
+                barrier_group_indices = []
+            post_output += f"{line}\n"
+        else:
+            barrier_group_indices.append(_get_barrier_index(line))
+    return post_output
```
rturrado, line 82: `ret`, or `result` are generic return value variable names that can fit good in this occasion.

```suggestion
    ret: str = ""
    barrier_group_indices: list[int] = []
    for line in output.split("\n"):
        if not line.startswith("barrier"):
            if barrier_group_indices:
                ret += _dump_barrier_group(barrier_group_indices)
                barrier_group_indices = []
            ret += f"{line}\n"
        else:
            barrier_group_indices.append(_get_barrier_index(line))
    return ret
```

opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -86,10 +86,15 @@ def _dump_barrier_group(indices: list[int]) -> str:
     return "barrier q[{}]\n".format(", ".join(map(str, indices)) if len(indices) != 0 else "")
 
 
+class CqasmV1ExporterParseError(Exception):
+    pass
```
rturrado, line 90: Can we move this class at the beginning of the file, before class _CQASMv1Creator?

elenbaasc: Done!

rturrado (approved):

PR #408, opened 2024-12-13
[CQT-233] Create bit mapping as property of the circuit

S-Linde:
# WARNING
This pull request has *BREAKING CHANGES*, do not merge without communicating this with end users.

### DIFF
--- a/opensquirrel/passes/exporter/quantify_scheduler_exporter.py
+++ b/opensquirrel/passes/exporter/quantify_scheduler_exporter.py
@@ -8,6 +8,7 @@ from opensquirrel.circuit import Circuit
 from opensquirrel.common import ATOL
 from opensquirrel.exceptions import ExporterError, UnsupportedGateError
 from opensquirrel.ir import BlochSphereRotation, ControlledGate, IRVisitor, MatrixGate, Measure, Reset
+from opensquirrel.utils import BitStringMapping
 
 try:
     import quantify_scheduler
@@ -30,11 +31,8 @@ class _ScheduleCreator(IRVisitor):
 
     def __init__(self, register_manager: RegisterManager) -> None:
         self.register_manager = register_manager
-        self.qubit_register_size = register_manager.get_qubit_register_size()
         self.qubit_register_name = register_manager.get_qubit_register_name()
-        self.bit_register_size = register_manager.get_bit_register_size()
-        self.acq_index_record = [0] * self.qubit_register_size
-        self.bit_string_mapping: list[tuple[None, None] | tuple[int, int]] = [(None, None)] * self.bit_register_size
+        self.bit_string_mapping = BitStringMapping()
         self.schedule = quantify_scheduler.Schedule("Exported OpenSquirrel circuit")
 
     def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
@@ -85,20 +83,16 @@ class _ScheduleCreator(IRVisitor):
         raise UnsupportedGateError(g)
 
     def visit_measure(self, g: Measure) -> None:
-        qubit_index = g.qubit.index
-        bit_index = g.bit.index
-        acq_index = self.acq_index_record[qubit_index]
-        self.bit_string_mapping[bit_index] = (acq_index, qubit_index)
+        self.bit_string_mapping.add_measure(g)
+        acq_channel, acq_index = self.bit_string_mapping.get_last_added_acq()
         self.schedule.add(
             quantify_scheduler_gates.Measure(
                 self._get_qubit_string(g.qubit),
-                acq_channel=qubit_index,
+                acq_channel=acq_channel,
                 acq_index=acq_index,
                 acq_protocol="ThresholdedAcquisition",
             )
         )
-        self.acq_index_record[qubit_index] += 1
-        return
 
     def visit_reset(self, g: Reset) -> Any:
         self.schedule.add(quantify_scheduler_gates.Reset(self._get_qubit_string(g.qubit)))
@@ -126,4 +120,4 @@ def export(circuit: Circuit) -> tuple[quantify_scheduler.Schedule, list[tuple[An
             "Decompose all gates to the Quantify-scheduler gate set first (rxy, rz, cnot, cz)"
         )
         raise ExporterError(msg) from e
-    return schedule_creator.schedule, schedule_creator.bit_string_mapping
+    return schedule_creator.schedule, schedule_creator.bit_string_mapping.to_export_format()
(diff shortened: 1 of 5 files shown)


rturrado:
Do we have a requirements document for this task? Can we explain what is a bit mapping in the description of the PR? What it is, why we need it, what it pretends to solve, what's the structure of this mapping?

opensquirrel/utils/bit_register_mapping.py
```diff
@@ -0,0 +1,73 @@
...
+                ...
+            },
+            ...
+            }
+        """
+        json_mapping: defaultdict[str, dict[str, int]] = defaultdict(dict)
+        for acq_channel, bit_indices in self._mapping.items():
+            for acq_idx, bit_idx in enumerate(bit_indices):
+                if self._bit_allocation[bit_idx] == acq_channel:
+                    json_mapping[str(acq_channel)][str(acq_idx)] = bit_idx
```
rturrado, line 72:
```suggestion
        json_mapping: defaultdict[str, dict[str, int]] = defaultdict(dict)
        for acq_channel, bit_indices in self._mapping.items():
            for acq_index, bit_index in enumerate(bit_indices):
                if self._bit_allocation[bit_index] == acq_channel:
                    json_mapping[str(acq_channel)][str(acq_index)] = bit_index
```

test/utils/test_bit_register_mapping.py
```diff
@@ -0,0 +1,71 @@
...
+    [
+        ([measure(0, 0)], {0: [0]}, {0: 0}),
+        ([measure(0, 0), measure(1, 0)], {0: [0], 1: [0]}, {0: 1}),
+        ([measure(0, 1), measure(0, 0)], {0: [1, 0]}, {0: 0, 1: 0}),
+    ],
+)
+def test_add_measure(
+    bit_string_mapping: BitStringMapping,
+    measurements: Iterable[Measure],
+    expected_mapping: dict[int, list[int]],
```
rturrado, line 28: Can we use a `BitStringMapping` for expected_mapping instead of a `dict[int, list[int]]`?

test/utils/test_bit_register_mapping.py
```diff
@@ -0,0 +1,71 @@
...
+    expected_bit_allocation: dict[int, int],
```
rturrado, line 29: Can we define a type for bit allocations? Would it be useful outside tests?

opensquirrel/utils/bit_register_mapping.py
```diff
@@ -0,0 +1,73 @@
...
+    from opensquirrel.ir import Measure
+
+
+class BitStringMapping:
+    """Utility class to create bitstringmappings for opensquirrel exporter passes."""
+
+    def __init__(self) -> None:
+        """Init of the ``BitStringMapping``."""
+        self._mapping: defaultdict[int, list[int]] = defaultdict(list)
+        self._bit_allocation: dict[int, int] = {}
```
rturrado, line 18: Maybe useful to define, at least, aliases for these mapping and bit allocation types?

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
```diff
@@ -85,26 +83,22 @@ def visit_controlled_gate(self, g: ControlledGate) -> None:
...
         )
-        self.acq_index_record[qubit_index] += 1
-        return
 
     def visit_reset(self, g: Reset) -> Any:
         self.schedule.add(quantify_scheduler_gates.Reset(self._get_qubit_string(g.qubit)))
 
 
-def export(circuit: Circuit) -> tuple[quantify_scheduler.Schedule, list[tuple[Any, Any]]]:
+def export(circuit: Circuit) -> tuple[quantify_scheduler.Schedule, dict[str, dict[str, int]]]:
```
rturrado, line 101: We should definitely use a type alias for `dict[str, dict[str, int]]]` (or more).

rturrado (commented):
Sorry, I wasn't even aware we had this bit string mappings in the Quantify Scheduler exporter.

It looks good to me, apart from:
- Variable names, comments...
- I would definitely use type aliases, or even classes, instead of using parameters and variables of complicated types, e.g., `dict[str, dict[...]]`.

opensquirrel/utils/bit_register_mapping.py
```diff
@@ -0,0 +1,73 @@
...
+        Args:
+            g: measurement to add the acquisition_channel and bitstring_index from.
+        """
+        acq_channel = g.qubit.index
+        bit_idx = g.bit.index
+
+        self._mapping[acq_channel].append(bit_idx)
+        self._bit_allocation[bit_idx] = acq_channel
+
+    def get_last_added_acq(self) -> tuple[int, int]:
```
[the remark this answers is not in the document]
S-Linde: Changed this to `get_last_added_acq_channel_and_index`, since it returns both the channel and index.

opensquirrel/utils/bit_register_mapping.py
S-Linde: Is `index` really more clear than `idx`? I think `idx` is a really common abbreviation, just like you would almost always say `iter` instead of `iteration`.

If you feel like it is unclear, because in your experience this abbreviation is less common than I think, than I'm all for changing it to `index`. Otherwise, I would say if we have two names that are equally clear in intend, we always choose the shorter one.

test/utils/test_bit_register_mapping.py
S-Linde: It is simply a mapping that maps classical bits to acquisition channels (qubit indices). In the output format each classical bit can only be used for one acquisition channel. This mapping is needed to see what we need should not export. It is only used for this purpose, so I'm not sure we need a type for it. It is only used in the tests and only makes sense with respect to this specific class.

rturrado: > It is simply a mapping that maps classical bits to acquisition channels (qubit indices).

Nobody knows that. Or has to know that. You won't know that in 3 months time. That's my whole point on using type aliases. If that line would read `expected_bit_allocation: MapBitQubit` or `map_bit_qubit_t` or whatever, it would be much more readable.

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
rturrado: Not in my opinion, no. Nothing stops us from giving a proper name to a `dict[str, dict[str, int]]`.

opensquirrel/utils/bit_register_mapping.py
rturrado: TBH, I would write `acquisition_index`.

test/utils/test_bit_register_mapping.py
rturrado: Arguably we shouldn't test private methods, private properties, just public methods. Anyway, my point is that any complex type exposed from the class should be simple to read. And aliases are good for that.

PR #411, opened 2024-12-18
[CQT-266] Routing checker pass

rares1609:
Added a Routing Checker pass


test/router/test_routing_checker.py
```diff
@@ -0,0 +1,25 @@
...
+from opensquirrel.passes.router.routing_checker import RoutingChecker
+
+
+@pytest.fixture(name="router")
+def router_fixture() -> RoutingChecker:
+    backend_connectivity_diagram = {0: [1, 2], 1: [0, 2, 3], 2: [0, 1, 4], 3: [1, 4], 4: [2, 3]}
+    return RoutingChecker(backend_connectivity_diagram)
+
+
+def test_routing_checker_possible_121_mapping(router: RoutingChecker) -> None:
```
elenbaasc, line 14:
```suggestion
def test_routing_checker_possible_1to1_mapping(router: RoutingChecker) -> None:
```

test/router/test_routing_checker.py
```diff
@@ -0,0 +1,25 @@
...
+    interactions = [(0, 1), (1, 2), (2, 4), (3, 4)]
+    try:
+        router.route(interactions)
+    except ValueError:
+        pytest.fail("route() raised ValueError")
```
elenbaasc, line 19: A circuit with interactions should be created with the `Circuitbuilder` and `assert` statements should be used to test (see tests on other passes f.i.).

test/test_integration.py
```diff
@@ -44,6 +45,10 @@ def test_spin_backend() -> None:
         """,
     )
 
+    # Check whether the above algorithm can be mapped to a dummy chip topology
+    connectivity = {0: [1, 2], 1: [0], 2: [0, 3], 3: [2]}
+
+    qc.route(interactions=[(0, 1), (2, 3), (0, 2)], router=RoutingChecker(connectivity))
```
elenbaasc, line 51: If the `route` methods in the various places (`Circuit.route`, `Router.route`, and `RoutingChecker.route`) is adjusted to accepting the IR as input, rather than the interactions, this could be simplified to:
 ```suggestion
    qc.route(router=RoutingChecker(connectivity))
```

**Review by elenbaasc — COMMENTED** [review]
Nice work, thanks for the effort!

Main point is that the qubit interactions should be inferred from the circuit, rather than providing them as user input. But close to be done I would say.

File: opensquirrel/passes/router/routing_checker.py [hunk 4]
```diff
@@ -0,0 +1,18 @@
...
+
+    def route(self, interactions: list[tuple[int, int]]) -> None:
+        non_executable_interactions = []
+
+        for q1, q2 in interactions:
+            if q2 not in self.backend_connectivity_diagram.get(q1, []):
+                non_executable_interactions.append((q1, q2))
+
+        if non_executable_interactions:
+            error_message = f"The qubit interactions: {non_executable_interactions} prevent a 121 mapping"
```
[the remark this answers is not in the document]
elenbaasc: We might want convert the non-executable interactions to a `set`, otherwise the same interactions may be displayed many times over.

elenbaasc (approved):
Looks good, thanks for making the changes!

When merging, make sure to use the option: **Squash and merge**

PR #412, opened 2024-12-19
[CQT-293] Remove generators

rturrado:
## Must read

This PR is mainly an investigation, although offering already a working implementation as a proof of concept. As such, I would advise to read first the [Confluence page](https://qutech-sd.atlassian.net/wiki/spaces/QW/pages/3700686850/Research+Remove+generators), and then use the PR as a companion to see the implementation changes.

## Changes

This PR includes also other changes and fixes (some of which may not be detailed in the Confluence page):
- Fix the `accept` implementation of the IR node classes.
- Make `try_match_replace_with_default` (old `try_name_anonymous_gate`) work with Rx, Ry, and Rz, e.g., Rx(pi/4) 4 times gives now Rx(pi), and not an anoymous Bloch.
- Make gate modifiers try naming an anonymous Bloch, e.g., `inv.inv.X` is now an X, not anonymous anymore.

### Change log

  - Change default instructions from callables to IR node classes.
  - Move default instructions to IR.
  - Remove generators from IR instructions.
  - Remove ANNOTATIONS_TO_TYPE_MAP.
  - Remove instruction decorators.
  - Remove instruction_library.py.


opensquirrel/passes/exporter/cqasmv1_exporter.py
```diff
@@ -33,38 +51,74 @@ def visit_float(self, f: SupportsFloat) -> str:
         f = Float(f)
         return f"{f.value:.{self.FLOAT_PRECISION}}"
 
+    def visit_gate(self, gate: Gate) -> None:
+        raise UnsupportedGateError(gate)
+
+    def visit_bloch_sphere_rotation(self, gate: Gate) -> None:
+        raise UnsupportedGateError(gate)
```
elenbaasc, line 58: Are both of these needed or only `visit_gate` as it would cover `visit_bloch_sphere_rotation`?

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
```diff
@@ -57,31 +70,44 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
 
         raise UnsupportedGateError(g)
 
+    def visit_bsr_without_params(self, g: BsrWithoutParams) -> None:
```
elenbaasc, line 73: To add to this refactor: also change `g` to `gate` in this module.

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
```diff
@@ -57,31 +70,44 @@ def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> None:
...
+                qT=self._get_qubit_string(g.target_qubit),
+            ),
+        )
+        return
+
+    def visit_cr(self, g: CR) -> None:
+        raise UnsupportedGateError(g)
 
+    def visit_crk(self, g: CRk) -> None:
         raise UnsupportedGateError(g)
```
elenbaasc, line 111: We have `UnsupportedGateError`, but shouldn't we also have unsupported _instruction_ errors for `init`, `wait`, and `barrier`, since the quantify-scheduler does not support these instructions (yet). Can also be added later, but then we should make a separate issue for it.

opensquirrel/reindexer/qubit_reindexer.py
```diff
@@ -58,20 +65,42 @@ def visit_wait(self, wait: Wait) -> Wait:
     def visit_bloch_sphere_rotation(self, g: BlochSphereRotation) -> BlochSphereRotation:
```
elenbaasc, line 65: Same goes for this module: to replace `g` with `gate`.

test/test_ir.py
```diff
@@ -263,9 +260,8 @@ def gate_fixture(self) -> MatrixGate:
     def test_array_like(self) -> None:
         gate = MatrixGate([[1, 0, 0, 0], [0, 1, 0, 0], [0, 0, 0, 1], [0, 0, 1, 0]], [0, 1])
         assert (
-            repr(gate) == "MatrixGate(qubits=[Qubit[0], Qubit[1]], "
-            "matrix=[[1.+0.j 0.+0.j 0.+0.j 0.+0.j]\n "
-            "[0.+0.j 1.+0.j 0.+0.j 0.+0.j]\n [0.+0.j 0.+0.j 0.+0.j 1.+0.j]\n [0.+0.j 0.+0.j 1.+0.j 0.+0.j]])"
+            repr(gate) == "MatrixGate(qubits=[Qubit[0], Qubit[1]], matrix=[[1.+0.j 0.+0.j 0.+0.j 0.+0.j] "
+            "[0.+0.j 1.+0.j 0.+0.j 0.+0.j] [0.+0.j 0.+0.j 0.+0.j 1.+0.j] [0.+0.j 0.+0.j 1.+0.j 0.+0.j]])"
```
elenbaasc, line 264: Isn't it better to represent these as a matrix (i.e. with the \n), instead of a line.

elenbaasc (commented):
Very nice refactor, thanks a lot for the effort!

Mostly some minor comments.

I see indeed the slight increase in overhead, but the clarity of how the ir is built and processed is improved a lot!

opensquirrel/ir.py
```diff
@@ -310,336 +346,354 @@ class Statement(IRNode, ABC):
...
+    def arguments(self) -> tuple[Expression, ...]:
+        pass
 
     @abstractmethod
     def get_qubit_operands(self) -> list[Qubit]:
         pass
 
+    @abstractmethod
+    def get_bit_operands(self) -> list[Bit]:
+        pass
```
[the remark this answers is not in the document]
rturrado: Think of it in this other way. In all the places where you want to generically treat an instruction (I can only find `CircuitBuilder::_check_out_of_bounds_access(self, instruction: Instruction)` for the time being), being able to get the bit operands of an instruction saves you from checking if it is a `Measure` instruction (also, tomorrow it could be `Measure` and something else).

Also, it doesn't look so bad, IMO, to ask an instruction for its qubits and its bits.

opensquirrel/ir.py
```diff
@@ -310,336 +346,354 @@ class Statement(IRNode, ABC):
...
```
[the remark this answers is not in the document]
rturrado: Thanks. I've tried removing them and it doesn't complain.

opensquirrel/ir.py
```diff
@@ -310,336 +346,354 @@ class Statement(IRNode, ABC):
...
     @property
-    def name(self) -> str:
-        return self.generator.__name__ if self.generator else "Unknown non-unitary instruction: " + self.__repr__()
+    @abstractmethod
+    def arguments(self) -> tuple[Expression, ...]:
+        pass
+
+    @property
+    def is_anonymous(self) -> bool:
+        return type(self) in [BlochSphereRotation, BsrWithoutParams, BsrWithAngleParam, MatrixGate, ControlledGate]
```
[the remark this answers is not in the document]
rturrado: All the default gates, default unitaries, default non-unitaries... in `default_instructions.py` are implemented as classes in `ir.py` (e.g., `class X`, `class CNOT`...), so they `type` will be `X`, `CNOT`... The line above just says: if you happen to examine a `Gate` of type `BlochSphereRotation`, `Matrix`, and so on, treat it like an anoymous gate. Otherwise you would find a `Gate` of type `X`, `CNOT`...

So I think it works, yet is not beautiful code. We could think of other options:
- Reimplement `is_anonymous` property with a positive logic, e.g., `return type(self) in [X, Y, Z...]`.
- Do not implement `is_anonymous` as a property of `Gate`, but as a helper function (or a set of helper functions) in `default_instructions.py` (e.g. `is_anonymous_gate`).
- Do not provide an implementation for `is_anonymous` and just substitute the check for a `gate.name in default_gate_set`. In this case I would implement a local function at least to write the check as `is_anonymous_gate(gate.name)`.

opensquirrel/ir.py
```diff
@@ -310,336 +346,354 @@ class Statement(IRNode, ABC):
...
     def accept(self, visitor: IRVisitor) -> Any:
-        visitor.visit_non_unitary(self)
-        return visitor.visit_init(self)
+        return visitor.visit_bloch_sphere_rotation(self)
 
     def get_qubit_operands(self) -> list[Qubit]:
         return [self.qubit]
 
+    def get_bit_operands(self) -> list[Bit]:
+        return []
```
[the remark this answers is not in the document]
rturrado: This is the implementation of `get_bit_operands` for the BSR subtree. So, at some point in the code, someone may have an Instruction and ask for the list of bits of that instruction. If the instruction happens to be a BSR, you want to return an empty list of bits.

opensquirrel/passes/exporter/cqasmv1_exporter.py
rturrado: Sorry, that was a typo, I've changed the gate type to BlochSphereRotation. Well seen.

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
rturrado: Done.

opensquirrel/passes/exporter/quantify_scheduler_exporter.py
rturrado: We can perfectly have a hierarchy or a set of exceptions, e.g., `UnsupportedInstructionError`, `UnsupportedGateError`, `UnsupportedNonGateError`, etc. But you have to add the new classes and  take a decision when you need them. For example, if at one point you one to raise an exception regarding `Measure` and you don't want to raise an `UnsupportedGateError`.

opensquirrel/reindexer/qubit_reindexer.py
rturrado: Done.

test/test_ir.py
rturrado: It may be, yes.

I changed the `repr_round` implementation to 1) do the conversion to string, and 2) remove the newlines. The advantage of the representation with this change is precisely that you have everything in one line. With the previous representation, the first row of the matrix would be in the first line, and then every other row in one line by itself.

You also have to think when you'll want this to use this method. Apart from tests, the only place where it's used is in the Writer. And the current implementation of the Writer removes the newlines when printing a Matrix:

```
            if "MatrixGate" in gate_name:
                # In the case of a MatrixGate the newlines should be removed from the array
                # such that the array is printed on a single line.
                gate_name = gate_name.replace("\n", "")
```

opensquirrel/ir.py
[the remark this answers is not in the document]
elenbaasc: Could we go for the first option and use the `default_instruction_set`? If it is in the set, it is not anonymous...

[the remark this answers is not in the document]
rturrado: I don't really like that because you would be using an external data structure to retrieve some information about yourself. The fact that the default instructions module already uses the IR module, where gates are defined, and thus that that implementation would introduce a circular dependency says something is not OK.

So either a gate should be able to define itself as anonymous without looking into an external data structure, or, if you want to define if a gate is anonymous based on a data structure, then you should probably go for a helper function in that default instructions module. I see the latter as a better option.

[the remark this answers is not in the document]
elenbaasc: Good point.

The gate should only know if it is anonymous, if it is part of the list of default instructions. So the helper function sounds like the better way to go.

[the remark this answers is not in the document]
rturrado: Done!

elenbaasc (approved):
Looks all good, thanks!

PR #425, opened 2025-01-14
[CQT-297] Check that composing bloch sphere rotations is done correctly

elenbaasc:
- Adjust `compose_bloch_sphere_rotations` to apply BSRs in stated order, _i.e._ first A then B.
- Update docstring to clarify order.
- Correct order input args where `compose_bloch_sphere_rotations` is used.
- Add tests where order of input args is tested specifically.


opensquirrel/passes/merger/general_merger.py
```diff
@@ -15,8 +15,16 @@
...
-    The first rotation is applied and then the second.
-    If the final Bloch sphere rotation is anonymous, we will try to match it to a default gate.
+    The first rotation (A) is applied and then the second (B):
+
+    As separate gates:
+        A q
+        B q
+
+    A linear operations:
+        B * A q
```
rturrado, line 25: Maybe clearer:
```suggestion
        (B * A) q
```

Or even `B*A q`.

opensquirrel/passes/merger/general_merger.py
```diff
@@ -15,8 +15,16 @@
...
+
+    If the final Bloch sphere rotation is anonymous, we try to match it to a named gate.
```
rturrado, line 27: The instruction library has default gates. If we remove generators, the named gate decorator will dissapper. So I think it's clearer to say `default` gate.

```suggestion
    If the final Bloch sphere rotation is anonymous, we try to match it to a default gate.
```

opensquirrel/passes/merger/single_qubit_gates_merger.py
```diff
@@ -25,7 +25,7 @@ def merge(self, ir: IR, qubit_register_size: int) -> None:
             instruction: Instruction = cast(Instruction, statement)
             if isinstance(instruction, BlochSphereRotation):
                 already_accumulated = accumulators_per_qubit[instruction.qubit]
-                composed = compose_bloch_sphere_rotations(instruction, already_accumulated)
+                composed = compose_bloch_sphere_rotations(already_accumulated, instruction)
```
rturrado, line 28: Nice!

opensquirrel/passes/merger/general_merger.py
```diff
@@ -40,7 +48,7 @@ def compose_bloch_sphere_rotations(bsr_a: BlochSphereRotation, bsr_b: BlochSpher
             * (
                 sin(bsr_a.angle / 2) * cos(bsr_b.angle / 2) * bsr_a.axis.value
                 + cos(bsr_a.angle / 2) * sin(bsr_b.angle / 2) * bsr_b.axis.value
-                + sin(bsr_a.angle / 2) * sin(bsr_b.angle / 2) * np.cross(bsr_a.axis, bsr_b.axis)
+                + sin(bsr_a.angle / 2) * sin(bsr_b.angle / 2) * np.cross(bsr_b.axis, bsr_a.axis)
```
rturrado, line 51: Good catch!

rturrado (commented):
Beautiful.

rturrado (approved):

PR #426, opened 2025-01-16
[CQT-277] Improve `is_anonymous` check for gates

GuyPuts:


opensquirrel/ir.py
```diff
@@ -521,7 +521,9 @@ def name(self) -> str:
 
     @property
     def is_anonymous(self) -> bool:
-        return self.generator is None
+        if self.generator:
+            return not (self.generator.__name__)
+        return True
```
rturrado, line 526: I would consider:

```
def _is_named_gate(self) -> bool:
   return self.generator and self.generator.__name__

@property
def is_anonymous(self) -> bool:
    return not self._is_named_gate()
```

GuyPuts: Good suggestion! I made _is_named_gate a property as well.

rturrado (approved):

opensquirrel/ir.py
```diff
@@ -519,9 +519,13 @@ def name(self) -> str:
             return self.generator.__name__
         return "Anonymous gate: " + self.__repr__()
 
+    @property
+    def _is_named_gate(self) -> bool:
+        return bool(self.generator and self.generator.__name__)
+
```
rturrado, line 525: Does it have to be a property? Does it make sense to have private properties?

rturrado (changes requested):
Sorry, I approved too fast.

opensquirrel/ir.py
rturrado: I see it OK!

opensquirrel/ir.py
```diff
@@ -520,12 +520,12 @@ def name(self) -> str:
         return "Anonymous gate: " + self.__repr__()
 
     @property
-    def _is_named_gate(self) -> bool:
+    def is_named_gate(self) -> bool:
         return bool(self.generator and self.generator.__name__)
```
rturrado, line 524: A couple of questions:
1. Do we need the cast to `bool` here? Isn't the result of `A and B` always bool?
2. Are the checks `self.generator` and `self.generator.__name__` correct? Shouldn't we check against `None` or something?

GuyPuts: I believe the checks were correct, but they did not return the right object, causing a MyPy error which required the use of the cast to bool. I've fixed both by doing this:
image

rturrado (approved):

PR #465, opened 2025-02-21
[CQT-307] Create SWAP2CZ replacer pass

elenbaasc:
- A SWAP2CZ replacer pass is created.
- The replacement is documented in the tutorial.
- Tests are written for it.

It is not added to the `test_integration.py`, because the backends don't use it currently.


test/decomposer/test_swap2cz_decomposer.py
```diff
@@ -0,0 +1,86 @@
...
+        (H(0), [H(0)]),
+        (Ry(0, 2.345), [Ry(0, 2.345)]),
+    ],
+    ids=["Hadamard", "rotation_gate"],
+)
+def test_ignores_1q_gates(decomposer: SWAP2CZDecomposer, gate: Gate, expected_result: list[Gate]) -> None:
+    check_gate_replacement(gate, expected_result)
+    assert decomposer.decompose(gate) == expected_result
+
+
```
juanboschero, line 33: I think it would be a good idea to test this with another strategy, like merge single qubit gates to check if the CZ decomposer still works and if the rys on the sides merge

juanboschero (changes requested):
Looking good, one small remark

test/decomposer/test_swap2cz_decomposer.py
elenbaasc: Do you mean testing something like the following circuit:

```
version 3.0

qubit[2] q

Ry(pi/2) q[1]
SWAP q[0], q[1]
Ry(-pi/2) q[1]
```

after applying `SWAP2CZDecomposer` and `SingleQubitGatesMerger`, becomes:

```
version 3.0

qubit[2] q

CZ q[0], q[1]
Ry(-pi/2) q[0]
Ry(pi/2) q[1]
CZ q[0], q[1]
Ry(pi/2) q[0]
Ry(-pi/2) q[1]
CZ q[0], q[1]
```

i.e., the single-qubit gates on the edges have been merged with the respective Ry rotations to become Identity gates and have been left out?

juanboschero: Exactly!

elenbaasc: That would more be a test for the merger functionality.

If I run the first circuit I get:
```
version 3.0

qubit[2] q

CZ q[0], q[1]
Y90 q[1]
BlochSphereRotation(qubit=Qubit[0], axis=[ 0. -1.  0.], angle=1.5708, phase=0.0)
CZ q[1], q[0]
Y90 q[0]
BlochSphereRotation(qubit=Qubit[1], axis=[ 0. -1.  0.], angle=1.5708, phase=0.0)
CZ q[0], q[1]
```

Which shows that the merger works as expected, but that the `try_match_replace_with_default` does not do a good job at recognizing the `mY90`.

I will make a ticket for this. But I don't think this test should be added to the tests for the `SWAP2CZDecomposer`.

juanboschero (approved):

PR #467, opened 2025-02-24
[CQT-338] Research two-qubit canonical representation

juanboschero:


opensquirrel/ir.py
```diff
@@ -678,6 +684,103 @@ def accept(self, visitor: IRVisitor) -> Any:
...
+    def to_matrix_gate(self) -> MatrixGate:
+        can_matrix = self.get_matrix()
+
+        return MatrixGate(can_matrix, operands=[self.control_qubit, self.target_qubit])
+
+    def is_identity(self) -> bool:
+        return np.allclose(self.get_matrix(), np.eye(2 ** len(self.operands)), atol=ATOL)
+
+
+class Can(CanonicalGate):
```
elenbaasc, line 770: The canonical form is insufficient at describing two-qubit gates, e.g., the CZ and CNOT gates have the same canonical form. That's why for the canonical representation (which will basically be our BlochSphereRotation representation analog for 2-qubit gates) needs be the KAK-decomposition, i.e. the 2-qubit interaction described by the canonical form and (at most) 4 single-qubit gates. But to be able to use this representation, we need to know for each 2-qubit gate in the standard gate library (i.e., CZ, CNOT, CR, CRk, SWAP), what the KAK-decomposition is. So this needs to be determined first, before we can implement the 2-qubit canonical representation.

elenbaasc (commented):
I imagine the canonical representation to be equal to the KAK-decomposition of a gate. That is, we use the KAK-decomposition as its primitive representation. However, we don't know the KAK-decomposition of the currently supported 2-qubit gates (i.e. CZ, CNOT, CR, CRk, and SWAP), nor do we have a general procedure to create this decomposition.

So, before we can continue with the implementation, we need to determine a general procedure to obtain the KAK-decomposition of arbitrary 2-qubit gates. I've come across the following sources, where this is described:
- [Optimal quantum circuits for general two-qubit gates (section V)](https://arxiv.org/pdf/quant-ph/0308006)
- [A geometric theory of non-local two-qubit operations (section III)](https://arxiv.org/pdf/quant-ph/0209120)
- [On gates (section 7.2)](https://threeplusone.com/pubs/on_gates.pdf)
- [Jupyter notebook example of KAK-decomposition](https://github.com/mpham26uchicago/laughing-umbrella/blob/main/background/Full%20Two%20Qubit%20KAK%20Implementation.ipynb)

PR #472, opened 2025-03-03
[CQT-346] Implement shortest path Routing pass

rares1609:


elenbaasc (commented):
Looks great! Thanks for the implementation.

My main comment is whether the IR needs to be copied for adding the SWAP gates. Other than that some minor comments.

Don't for get to add an entry to the Changelog ;)!

test/router/test_shortest_path_router.py
```diff
@@ -0,0 +1,99 @@
...
+def test_medium_case(router2: ShortestPathRouter, circuit2: Circuit) -> None:
+    new_ir = router2.route(circuit2.ir)
+    swap_count = sum(1 for statement in new_ir.statements if isinstance(statement, SWAP))
+    assert swap_count == 8
+
+
+def test_complex_case(router3: ShortestPathRouter, circuit3: Circuit) -> None:
+    new_ir = router3.route(circuit3.ir)
+    swap_count = sum(1 for statement in new_ir.statements if isinstance(statement, SWAP))
+    assert swap_count == 15
```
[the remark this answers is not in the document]
rares1609: Combined tests

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -0,0 +1,35 @@
+import networkx as nx
+
+from opensquirrel.ir import IR, SWAP, Gate, Qubit
+from opensquirrel.passes.router import Router
+
+
+class ShortestPathRouter(Router):
+    def __init__(self, connectivity_scheme: dict[int, list[int]]) -> None:
```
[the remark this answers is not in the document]
rares1609: Done

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -0,0 +1,35 @@
...
+        to make the circuit executable on the given hardware connectivity scheme.
+
+        Args:
+            ir (IR): The intermediate representation of the circuit to be routed.
+
+        Returns:
+            new_ir (IR): The new intermediate representation with the additional SWAP gates.
+        """
+        graph = nx.Graph(self.connectivity_scheme)
+        new_ir = IR()
```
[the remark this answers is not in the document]
rares1609: I restructured the function (the loop mainly), and now no new ir is made. The tests behave the same as before.

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def route(self, ir: IR) -> IR:
+        """
+        Routes the circuit by inserting SWAP gates to make it executable given the hardware connectivity.
+        Args:
+            ir: The intermediate representation of the circuit.
+        Returns:
+            The intermediate representation of the routed circuit (including the additional SWAP gates).
+        """
+        graph = nx.Graph({int(k): [int(v) for v in values] for k, values in self.connectivity.items()})
```
elenbaasc, line 19: The values should be integers already, right? So no need for a conversion. Also good to specify what the keys and values are.

```suggestion
        graph_data = {int(start_node): end_nodes for start_node, end_nodes in self.connectivity.items()}
        graph = nx.Graph(graph_data)
```

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -0,0 +1,34 @@
...
+        instruction_counter = 0
```
elenbaasc, line 20: Instruction is only a specific kind of statement and you are going through a list of statements. I would use a 'counter' if I would actually what to know the sum total of something (at some point), but here it is used as an index.

```suggestion
        statement_index = 0
```

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -0,0 +1,34 @@
...
+            if isinstance(statement, Gate) and len(statement.get_qubit_operands()) == 2:
+                q0, q1 = statement.get_qubit_operands()
+                if not graph.has_edge(q0.index, q1.index):
+                    try:
+                        shortest_path = nx.shortest_path(graph, source=q0.index, target=q1.index)
+                        for start_qubit_index, end_qubit_index in zip(shortest_path[:-1], shortest_path[1:]):
+                            ir.statements.insert(instruction_counter, SWAP(start_qubit_index, end_qubit_index))
+                            instruction_counter += 1
+                    except nx.NetworkXNoPath:
+                        print(f"No routing path available between qubit {q0.index} and qubit {q1.index}")  # noqa: T201
```
elenbaasc, line 32: We should create a custom exception in `opensquirrel.exceptions` for this with the message and then create something like:
```
try:
   ...
except NoRoutingPathError as e:
   msg = f"No routing path available between qubit {q0.index} and qubit {q1.index}"
   raise NoRoutingPathError(msg)
```

elenbaasc (commented):
Thanks for the work! Happy that the IR doesn't need to be copied :)

Some extra comments.

I changed the handling of the fixtures a bit so that we don't have to rely on an extra dependency.

opensquirrel/passes/router/shortest_path_router.py
rares1609: Done

PR #473, opened 2025-03-03
[CQT-308] Create CZ Decomposer

juanboschero:
Analog to the CNOTDecomposer, we require a CZDecomposer. The former is used to decompose arbitrary 2-qubit gates into a decomposition of single-qubit gates and CNOTs. However, CZ is a more general native gate (i.e. the CNOT is decomposed to CZ + single-qubit gates), so it would be better to decompose arbitrary 2-qubit gates already to single-qubit gates and CZs. Accordingly, we do not require decomposition from CNOT to CZ anymore.


rares1609 (commented):
The issue looks good! I would say its almost good for merging. 

The only comments I would have are the following:

- Rename `test_cz_decomposed.py` to `test_cz_decomposer.py`
- I have not seen any changes done to `test_integration.py`, and on jira it says that the pass should be added to the tests there.
- The description on the jira issue also mentioned that the pass should be documented in the tutorial. However, I suppose I will do that when I write the documentation for issue [CQT-286](https://qutech-sd.atlassian.net/jira/software/c/projects/CQT/boards/18?selectedIssue=CQT-286)? So, I think it can be merged without this requirement.

rares1609 (commented):
It looks great! 

The only thing which I am not sure about, is whether the pass needs to be added to `test_integration.py` or not (as it says on the Jira ticket). Furthermore, on the ticket its also stated that the `CNOT2CZDecomposer` is no longer needed, but it is still present in `test_integration.py`.

If this is not the case anymore, then I would say its good to merge.

elenbaasc:
> It looks great!
> 
> The only thing which I am not sure about, is whether the pass needs to be added to `test_integration.py` or not (as it says on the Jira ticket). Furthermore, on the ticket its also stated that the `CNOT2CZDecomposer` is no longer needed, but it is still present in `test_integration.py`.
> 
> If this is not the case anymore, then I would say its good to merge.

Good point! I actually think now that we should only change the `test_integration.py` if the backends have made these changes to the compile configuration. So we will need to keep it as is for now. (sorry for the confusion, the I should not put that as part of the acceptance criteria :P)

rares1609 (approved):

PR #483, opened 2025-03-17
[CQT-354] Integrate libQASM 1.1.0

elenbaasc:
- Change `Parser` to accept instruction parameters (instead of a single parameter)
- Update `pyproject.toml` and `poetry.lock` file to use libQASM 1.1.0
- Update Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 * **Removed** for now removed features.
 
 
-## [0.3.2] - [ xxxx-yy-zz ]
+## [0.4.0] - [ xxxx-yy-zz ]
 
 ### Added
 
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Update libQASM to 1.1.0 (from 0.6.9)
 - Refactor: removed generators
 - Changed the `RoutingChecker` pass to a `RoutingValidator` pass
 
--- a/opensquirrel/parser/libqasm/parser.py
+++ b/opensquirrel/parser/libqasm/parser.py
@@ -109,13 +109,13 @@ class Parser:
         return ret
 
     @classmethod
-    def _get_named_gate_param(cls, gate: cqasm.semantic.Gate) -> Any:
-        """Get the parameter of a named gate.
+    def _get_named_gate_parameters(cls, gate: cqasm.semantic.Gate) -> Any:
+        """Get the parameters of a named gate.
         Notice the input gate can be a composition of gate modifiers acting on a named gate.
         """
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
+- Update libQASM to 1.1.0 (from 0.6.9)
```
rturrado, line 22: Previous CHANGELOG message was: `libQASM 0.6.9 integrated (updated from 0.6.7)`.
Can we decide for one or other sentence and use it in the two places?

opensquirrel/parser/libqasm/parser.py
```diff
@@ -126,12 +126,12 @@ def _get_expanded_instruction_args(self, instruction: cqasm.semantic.Instruction
         """
         expanded_args = self._get_instruction_operands(instruction)
         if isinstance(instruction, cqasm.semantic.GateInstruction):
-            gate_parameter = self._get_named_gate_param(instruction.gate)
+            gate_parameters = self._get_named_gate_parameters(instruction.gate)
         else:
-            gate_parameter = self._ast_literal_to_ir_literal(instruction.parameter)
-        if gate_parameter:
+            gate_parameters = [self._ast_literal_to_ir_literal(parameter) for parameter in instruction.parameters]
+        if gate_parameters:
```
rturrado, line 132:
```suggestion
        if len(gate_parameters) > 0:
```

Before, `gate_parameter` was an optional. Now, it will always be a list, and it could be empty but never `None` (so `if gate_parameters:` check would always pass).

opensquirrel/parser/libqasm/parser.py
```diff
@@ -170,10 +170,11 @@ def _get_gate_generator(self, instruction: cqasm.semantic.GateInstruction) -> Ca
             if gate_name == "inv":
                 return InverseGateModifier(modified_gate_generator)
             if gate_name == "pow":
-                return PowerGateModifier(instruction.gate.parameter.value, modified_gate_generator)
+                return PowerGateModifier(instruction.gate.parameters[0].value, modified_gate_generator)
```
rturrado, line 173:
```suggestion
                gate = instruction.gate
                exponent = gate.parameters[0]
                return PowerGateModifier(instruction.gate.parameters[0].value, modified_gate_generator)
```

Just a suggestion. It maybe clarifies the code.

rturrado (commented):
Nice! Thanks for picking this up!

opensquirrel/parser/libqasm/parser.py
rturrado: OK, my bad, then just leave it as it is. Thanks for the explanation.

rturrado (approved):

PR #488, opened 2025-03-20
Updates tests and connectivity for the HQ/2 setup

TimVroomans:


test/test_integration.py
```diff
@@ -239,30 +240,21 @@ def test_hectoqubit_backend() -> None:
...
-        "mY90",
-        "Z",
-        "S",
-        "Sdag",
-        "T",
-        "Tdag",
         "Rx",
         "Ry",
         "Rz",
-        "CNOT",
```
elenbaasc, line 264: I understand that these gates are not native to the backend, but they are the gates that are supported by the quantify-scheduler Exporter. So, when validation is done after the decomposition stage, the compiler needs to understand which gates can be _translated_ (not necessarily which are natively supported).

We could also call them the primitive gate set to avoid confusion.

elenbaasc (commented):
Thanks for making the changes!

Some minor comments.

I have run the tests, and they run fine btw.

.gitignore
```diff
@@ -161,3 +161,7 @@ cython_debug/
 #  and can be added to the global gitignore or merged into this file.  For a more nuclear
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
+example.png
+
+# Pre Commit
+.pre-commit-config.yaml
```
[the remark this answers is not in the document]
TimVroomans: I see, I'll remove the changes to `.gitignore`

test/test_integration.py
TimVroomans: The reason for this change was that I would like to make sure that all operations for the Hectoqubit setup are compiled into Rxy, Rz and CZ operations.

I do now realise that this is not the place to check for it. These are just the gates that have a one-to-one mapping to an operation in Quantify Scheduler and those that are explicitly defined in my Quantify Scheduler [compiler configuration](https://quantify-os.org/docs/quantify-scheduler/dev/autoapi/quantify_scheduler/backends/graph_compilation/index.html#quantify_scheduler.backends.graph_compilation.CompilationConfig).

For that reason I will put all the single-qubit gates back in, but leave the CNOT out. While it does have a one-to-one mapping to an operation in Quantify Scheduler, my compiler configuration does not have an explicit definition of a CNOT gate.

PR #493, opened 2025-03-28
[CQT-336] Add Rn gate to compiler

elenbaasc:
- Add Rn gate to codebase
- Update `_get_expanded_instruction_args` in `parser.py` to accept lists of parameters
- Add Rn gate to tests (including a new test module to test supported instructions, _i.e._, `test_instructions.py`)


opensquirrel/writer/writer.py
```diff
@@ -73,6 +74,15 @@ def visit_bsr_no_params(self, gate: BsrNoParams) -> None:
...
 
+    def visit_bsr_full_params(self, gate: BsrFullParams) -> None:
+        nx = gate.nx.accept(self)
+        ny = gate.ny.accept(self)
+        nz = gate.nz.accept(self)
+        theta_argument = gate.theta.accept(self)
+        phi_argument = gate.phi.accept(self)
+        qubit_operand = gate.qubit.accept(self)
+        self.output += f"{gate.name}({nx}, {ny}, {nz}, {theta_argument}, {phi_argument}) {qubit_operand}\n"
+
```
rares1609, line 85: The parameters from `BsrFullParams` match the ones in the description on Jira, so again, this looks alright.

opensquirrel/ir.py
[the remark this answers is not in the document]
elenbaasc: Thanks! If you have no comments or require any further changes you can always immediately 'Approve' 😄

test/merger/test_single_qubit_gates_merger.py
```diff
@@ -61,7 +61,7 @@ def test_merge_different_qubits(merger: SingleQubitGatesMerger) -> None:
     circuit = builder1.to_circuit()
 
     builder2 = CircuitBuilder(4)
-    builder2.ir.add_gate(BlochSphereRotation(0, axis=(1, 0, 1), angle=math.pi))  # this is Hadamard with 0 phase
+    builder2.ir.add_gate(Rn(0, 1, 0, 1, math.pi, 0))  # this is Hadamard with 0 phase
```
rares1609, line 64: This is more of a question to understand the declaration of Rn gates. Is this gate here applied to qubit 0 in the circuit?

test/test_instructions.py
rares1609, line 1: All the instructions in the cQASM string seem to match those in the assert statement.

The question I would have, I suppose that for the `init`, `reset`, `wait` and `measure` statements, for instance when one does `init q`, that is applied to all qubits in the circuit? (e.g. `init q` => `init q[0]` and `init q[1]`)

rares1609 (commented):
This issue looks great! The Rn gate was added to the IR of OpenSquirrel and other relevant files, and it seems to follow the characteristics mentioned in [the documentation](https://qutech-sd.atlassian.net/wiki/spaces/QW/pages/3667853316/Requirements#Rn-(arbitrary-Bloch-sphere-rotation)).

There is only one thing left to add: updating the changelog file :).

test/merger/test_single_qubit_gates_merger.py
elenbaasc: Yes, indeed. The first parameter of the Rn gate (as a callable in OpenSquirrel) is the qubit.

test/test_instructions.py
elenbaasc: Yes, if `q` is defined as an array of qubits (_e.g._, `qubit[2] q`) then if one applies an instruction on that array, the instruction will be applied to all qubits in that array (Note that you can defined multiple arrays in a circuit, so it will no necessarily be applied to all qubits in the circuit.

rares1609: I see

rares1609 (dismissed):

PR #496, opened 2025-04-01
[CQT-355] Add support for asm declarations

elenbaasc:
- Implement asm declaration (as statement in `IR`).
- Integrate in `Parser`, `CircuitBuilder`, `Writer`.
- Add check in `SingleQubitGatesMerger` (to not merge across asm declarations).
- Add tests.
- Add entry in Changelog.


opensquirrel/ir.py
```diff
@@ -406,6 +452,21 @@ def is_identity(self) -> bool:
         pass
 
 
+class AsmDeclaration(Statement):
```
juanboschero, line 455: Can you provide docstrings? It is not immediately clear as to what an "ASM" is for a user looking at the general IR

opensquirrel/ir.py
```diff
@@ -959,6 +1020,9 @@ class IR:
     def __init__(self) -> None:
         self.statements: list[Statement] = []
 
+    def add_asm_declaration(self, asm_declaration: AsmDeclaration) -> None:
```
juanboschero, line 1023: Out of curiosity, why do you add an asm declaration only to the BlochSphereRotation?

juanboschero (changes requested):

opensquirrel/ir.py
elenbaasc: I have added some doc strings to the `AsmDeclaration` class. This is the first then of IR Statements that has a docstring. The user will/should, in principle, not have access to the IR nodes, but I agree that it would be good that all classes/methods (w/ the exception of helper methods) have doc strings.

opensquirrel/ir.py
elenbaasc: I'm not sure if I understand your question. An asm declaration is not added to a BlochSpehereRotation... There are now 2 kinds of `Statement`: `Instruction` statements and `AsmDeclaration` statements. `Instruction` statements then (ultimately) divide into gates and non-unitaries. Asm declarations, gates, and non-unitaries are all individually added to the IR; hence the `add_asm_declaration`, `add_gate`, and `add_non_unitary` methods in the `IR` class.

opensquirrel/circuit_builder.py
```diff
@@ -73,11 +73,19 @@ def _check_out_of_bounds_access(self, instruction: Instruction) -> None:
...
+    def _add_statement(self, attr: str, *args: Any) -> Self:
+        if attr == "asm":
+            try:
+                asm_declaration = AsmDeclaration(*args)
+                self.ir.add_asm_declaration(asm_declaration)
+            except TypeError:
+                msg = f"trying to build '{attr}' with the wrong number or type of arguments: '{args}'"
+                raise TypeError(msg) from None
+            return self
+
```
rares1609, line 85: This looks very clear to me

opensquirrel/ir.py
```diff
@@ -352,6 +398,28 @@ class Statement(IRNode, ABC):
...
+    ) -> None:
+        self.backend_name = String(backend_name)
+        self.backend_code = String(backend_code)
+        Statement.__init__(self)
+
+    def accept(self, visitor: IRVisitor) -> Any:
+        visitor.visit_statement(self)
+        return visitor.visit_asm_declaration(self)
+
+
```
rares1609, line 422: This looks in order. The `AsmDeclaration` class has the `backend_name` and `backend_code` attributes, as mentioned in the Jira ticket.

opensquirrel/ir.py
```diff
@@ -117,6 +123,46 @@ class Expression(IRNode, ABC):
     pass
 
 
+@runtime_checkable
+class SupportsStr(Protocol):
+    def __str__(self) -> str: ...
```
rares1609, line 128: I suppose this protocol is needed because there is no `SupportsStr` in `typing`?

rares1609 (commented):
This issue looks good! The necessary additions have been made to `circuit_builder.py` and `ir.py`, and the `AsmDeclaration` class has the 2 attributes mentioned in the ticket. Furthermore, new tests have been added for this feature.

The only thing that seems to be missing is the `test_rydberg_backend` that should be added to `test_integration.py` (according to the jira ticket).

opensquirrel/ir.py
elenbaasc: Yes, indeed!

rares1609 (approved):
Looks good to go!

PR #497, opened 2025-04-01
[CQT-357] Change name of native gate validator to primitive gate validator

elenbaasc:
- Changed use of 'native' to 'primitive' through code base.


opensquirrel/passes/validator/primitive_gate_validator.py
```diff
@@ -2,25 +2,27 @@
...
         Raises:
-            ValueError: If any unitary gate in the circuit is not part of the native gate set.
+            ValueError: If any unitary gate in the circuit is not part of the primitive gate set.
         """
-        gates_not_in_native_gate_set = [
+        gates_not_in_primitive_gate_set = [
             statement.name
             for statement in ir.statements
-            if isinstance(statement, Unitary) and statement.name not in self.native_gate_set
+            if isinstance(statement, Unitary) and statement.name not in self.primitive_gate_set
```
rares1609, line 22: In the Jira ticket, it is mentioned that the `PrimitiveGateChecker` must also check non-unitary gates. Therefore, maybe it should be:

`if isinstance(statement, Gate) and statement.name not in self.primitive_gate_set`

rares1609 (commented):
This looks almost good to merge! I was not able to find a single occurrence of the word "native".

The only thing I would have to comment, is that the Jira ticket states that "Non-unitaries are also checked by the Validator". However, at the moment, the `PrimitiveGateValidator` still only checks unitary gates.

opensquirrel/passes/validator/primitive_gate_validator.py
elenbaasc: Great catch! You're right this should be changed 😝

rares1609 (approved):

PR #531, opened 2025-05-23
[CQT-381] The Routing validator should skip asm declarations

elenbaasc:
Fix ignore asm declaration in RoutingValidator.
Add test to ignore asm declarations.
Add entry to Changelog.

### DIFF
--- a/opensquirrel/passes/validator/routing_validator.py
+++ b/opensquirrel/passes/validator/routing_validator.py
@@ -1,6 +1,6 @@
 from typing import Any, cast
 
-from opensquirrel.ir import IR, Instruction, Qubit
+from opensquirrel.ir import IR, AsmDeclaration, Instruction, Qubit
 from opensquirrel.passes.validator import Validator
 
 
@@ -21,6 +21,8 @@ class RoutingValidator(Validator):
         """
         non_executable_interactions = []
         for statement in ir.statements:
+            if isinstance(statement, AsmDeclaration):
+                continue
             instruction: Instruction = cast("Instruction", statement)
             args = instruction.arguments
             if args and len(args) > 1 and all(isinstance(arg, Qubit) for arg in args):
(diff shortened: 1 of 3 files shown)


opensquirrel/passes/validator/routing_validator.py
rares1609, line 1: This all makes sense. However, would it not be more specifically to have the code skip iterations where there is specifically an `AsmDeclaration`?

opensquirrel/passes/validator/routing_validator.py
```diff
@@ -21,8 +21,9 @@ def validate(self, ir: IR) -> None:
         """
         non_executable_interactions = []
         for statement in ir.statements:
-            instruction: Instruction = cast("Instruction", statement)
-            args = instruction.arguments
+            if not isinstance(statement, Instruction):
+                continue
```
rares1609, line 25:
```suggestion
            if not isinstance(statement, AsmDeclaration):
                continue
```
And obviously `from opensquirrel.ir import AsmDeclaration`

test/validator/test_routing_validator.py
rares1609, line 1: This looks great! An extra test is added to see whether Asm Declarations are skipped, and since the test passes I suppose it works as it should.

rares1609 (approved):
This pull request looks in order :). My only comment is whether it would be more preferrable to have the code specifically skip Asm Declarations in `routing_validator.py`. 

I will however mark this as approved in case you think its not necessary, so that you can merge it either way.

opensquirrel/passes/validator/routing_validator.py
elenbaasc: No, because in the future we might have other statements, that are not instructions, that also need to be skipped. What is important here is not that it is not an asm declaration, but that it is an instruction which has qubit arguments.

opensquirrel/passes/validator/routing_validator.py
rares1609: Alright! I also just notices the suggestion I made was wrong to begin with :) so sorry about that

PR #534, opened 2025-05-26
[CQT-364] Revise and test known gate identification after merging

elenbaasc:
- Updated gate identification to check for BSR equality.
- Added (absolute) phase check to BSR equality.
- Added correct phases to standard gates.
- Updated tests in accordance with new equality check (now with phase included)


rares1609 (approved):
Looks good! I'd say its good to be merged.👍

PR #563, opened 2025-07-09
[CQT-243] Implementation of global phase

juanboschero:
- Changed circuit class
- Added PhaseMap class
- Changed internal decomposer logic
- Added helper functions
- Add tests


opensquirrel/passes/decomposer/general_decomposer.py
```diff
@@ -40,25 +61,42 @@ def check_gate_replacement(gate: Gate, replacement_gates: Iterable[Gate]) -> Non
...
+        if len(gate_qubit_indices) > 1:
+            relative_phase = float(
+                np.real(
+                    circuit.phase_map.get_qubit_phase(qubit_list[1]) - circuit.phase_map.get_qubit_phase(qubit_list[0])
+                )
+            )
+            if abs(relative_phase) > ATOL:
+                list(replacement_gates).append(Rz(gate.get_qubit_operands()[0], Float(-1 * relative_phase)))
+
+    return list(replacement_gates)
```
elenbaasc, line 79:
```suggestion
        if len(qubits) > 1:
            relative_phase = float(
                np.real(
                    circuit.phase_map.get_qubit_phase(qubits[0]) - circuit.phase_map.get_qubit_phase(qubits[1])
                )
            )
            if abs(relative_phase) > ATOL:
                list(replacement_gates).append(Rz(qubits[0], Float(relative_phase)))

    return list(replacement_gates)
```

opensquirrel/phasemap.py
```diff
@@ -0,0 +1,24 @@
...
+class PhaseMap:
+    def __init__(self, phase_map: NDArray[np.complex128]) -> None:
+        """Initialize a PhaseMap object."""
+        self.qubit_phase_map = phase_map
+
+    def __contains__(self, qubit: QubitLike) -> bool:
+        """Checks if qubit is in the phase map."""
+        return qubit in self.qubit_phase_map
+
+    def add_qubit_phase(self, qubit: QubitLike, phase: np.complex128) -> None:
```
elenbaasc, line 16:
```suggestion
    def set_qubit_phase(self, qubit: QubitLike, phase: np.complex128) -> None:
```

opensquirrel/phasemap.py
```diff
@@ -0,0 +1,24 @@
...
+        from opensquirrel.ir import Qubit
```
elenbaasc, line 17: These should be imported at module level.

tests/integration/test_integration.py
```diff
@@ -780,3 +782,60 @@ def test_rydberg_backend() -> None:
 X q[8]
 """
     )
+
+
+def test_integration_global_phase() -> None:
```
elenbaasc, line 787: This should be part of a separate `test_phase_map.py` module. The `test_integration.py` is there to test for specific backend.

elenbaasc (commented):
Thanks for the effort!

I think that the PhaseMap could simply track the phase angle (so no need for an array of complex values).

Also, the phase map still needs a separate test module.

And don't forget to add an entry to the Changelog.

opensquirrel/ir.py
```diff
@@ -275,6 +275,12 @@ def __init__(self, index: QubitLike) -> None:
             msg = "index must be a QubitLike"
             raise TypeError(msg)
 
+    def __eq__(self, other: Any) -> bool:
+        """Compare two qubits."""
+        if not isinstance(other, Qubit):
+            return False
+        return self.__hash__() == other.__hash__()
```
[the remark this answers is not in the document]
juanboschero: I think it's best if we don't equate qubitlike to qubit. In our internal code, we serialize qubitlike objects to qubits when we wish to use them as such.

opensquirrel/phase_map.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __contains__(self, qubit: QubitLike) -> bool:
+        """Checks if qubit is in the phase map."""
+        return qubit in self.qubit_phase_map
+
+    def add_qubit_phase(self, qubit: QubitLike, phase: np.float64) -> None:
+        self.qubit_phase_map[Qubit(qubit).index] += phase
+
+    def get_qubit_phase(self, qubit: QubitLike) -> np.float64:
+        return np.float64(self.qubit_phase_map[Qubit(qubit).index])
```
elenbaasc, line 20:
```suggestion
    def add_qubit_phase(self, qubit: QubitLike, phase: float) -> None:
        self.qubit_phase_map[Qubit(qubit).index] += phase

    def get_qubit_phase(self, qubit: QubitLike) -> float:
        return self.qubit_phase_map[Qubit(qubit).index]
```

elenbaasc (commented):
Thanks for making the adjustments!

Mainly the tests need to be updated. It is not clear what the function is of the phase map (at least not from the tests). There needs to be a test that explicitly shows that the compiled circuit has been adjusted/corrected for phase differences between interacting qubits. Current tests don't show/test this.

Other than that, there are still some minor changes and some comments still left from the previous review round.

tests/test_phase_map.py
```diff
@@ -0,0 +1,77 @@
...
+import math
+
+import numpy as np
+
+from opensquirrel import Circuit, CircuitBuilder
+from opensquirrel.passes.decomposer import CNOT2CZDecomposer, CNOTDecomposer, McKayDecomposer, XYXDecomposer
+from opensquirrel.passes.merger import SingleQubitGatesMerger
+
+
+def test_integration_global_phase() -> None:
```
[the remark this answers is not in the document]
juanboschero: I made it a bit simpler and explicitly left it as a high level test

PR #567, opened 2025-07-18
[CQT-393] Check if tox can be removed from default dependencies

elenbaasc:
- Update poetry.lock file.
- Remove tox dependency from defaults.


davidbakker123 (approved):
`tox` is removed from the default dependencies and this doesn't lead to any problems. This makes sense, since it is only used during development.

PR #596, opened 2025-08-18
[CQT-401] OpenSquirrel's version is not exposed 

elenbaasc:
- Adjusted `opensquirrel/__init__.py` to expose version.


davidbakker123 (approved):
Looks good. Is it worth writing a test for this?

PR #597, opened 2025-08-18
[CQT-379] Add mike User documentation deployment to OpenSquirrel actions

elenbaasc:
- Create `docs.yml` for docs deployment using `mike`.
- Add `mike` to docs dependencies.


.github/workflows/docs.yml
```diff
@@ -0,0 +1,58 @@
...
+        if: github.ref == 'refs/heads/develop'
+        run: |
+            uv run mike deploy --push latest
+            uv run mike set-default --push latest
+
+      - name: Upload and tag as git tag
+        if: github.event_name == 'release' && github.event.action == 'created'
+        run: |
+          uv run mike deploy --push ${{ github.ref_name }}
+          uv run mike set-default --push latest
```
davidbakker123, line 58: `uv run mike set-default --push latest` will always run, so you can make it a separate step.

elenbaasc: Oh, nice catch 😃

davidbakker123: Sure also fine. You could still add that condition in the separate step, but I will leave you alone haha :P

davidbakker123 (approved):

PR #598, opened 2025-08-18
[CQT-395] Transfer to using uv instead of poetry for dependency managment

davidbakker123:
OpenSquirrel uses poetry for dependency management, but a leaner, more flexible alternative is available, i.e., [uv](https://docs.astral.sh/uv/).


CONTRIBUTING.md
davidbakker123, line 1: I tried to update file as best as I could, but I'm not sure if this is correct. Also, I don't us PyCharm, so that part was just a guess. This file seems to already be outdated anyway...

elenbaasc:
The export dependencies are not available for macOS arm architecture, so I have disabled them for those platforms.

For static analysis and unit testing, only the _default_, _dev_, and _export_ dependencies are required. The command `uv sync --group export` installs only dependencies from those groups (the _default_ and _dev_ groups are installed by default).

I guess this is ready to be merged :)

elenbaasc (approved):
Good to go!

Whether the `release` workflow runs properly, we'll see once we make a new release.

PR #601, opened 2025-08-20
[CQT-402] Images and elements not visible in online user docs

elenbaasc:
Update doc dependencies.
Change reference to images.
Update git ignore.


rares1609 (approved):
Looks in order! So ready to be merged.

PR #612, opened 2025-09-08
[CQT-341] Implement two qubit canonical representation as two qubit gate semantic

davidbakker123:


elenbaasc (commented):
Looks good!

Was there a particular reason why you moved the `Axis` and `CanonicalAxis` from the `Expression`s?

elenbaasc (dismissed):
Looks all good to me! Thanks for making the changes :)

rares1609 (approved):

PR #615, opened 2025-09-12
[CQT-410] Move pytest options from pyproject.toml to tox.ini to enable debugging

elenbaasc:
- Move pytest config from pyproject.toml to tox.ini.


rares1609 (approved):
Its good to go! The pytest options are moved to `tox.ini` as stated in the ticket description 👍.

PR #626, opened 2025-10-10
[CQT-361] - 2-Proper order schedule

elenbaasc:
- Processing control instructions (barrier and wait) added to the quantify-scheduler exporter.
- Introduced the `_Scheduler` visitor and `OperationRecord`
- Implementation performs As Late As Possible (ALAP) scheduling: the Scheduler processes the control instructions in reverse.
- Tests added.

**TODO**: Update Changelog after OpenSquirrel 0.7.0 release.

**_NOTE:_** This branch has additional base development branches that can be removed after merging to develop. These branches are:
- [CQT-361-Add-ALAP-scheduling-to-quantify-scheduler-Exporter](https://github.com/QuTech-Delft/OpenSquirrel/tree/CQT-361-Add-ALAP-scheduling-to-quantify-scheduler-Exporter)
- [CQT-257-Integrate-wait-and-barrier-directives-in-quantify-scheduler-Exporter](https://github.com/QuTech-Delft/OpenSquirrel/tree/CQT-257-Integrate-wait-and-barrier-directives-in-quantify-scheduler-Exporter)


tests/passes/exporter/test_quantify_scheduler_exporter.py
rares1609: Another interesting test would be, what if we have a simple circuit in which `wait` is applied multiple times on the same qubit. it could look something like:

```python
def test_multiple_waits_on_same_qubit(qs_is_installed: bool):
    if qs_is_installed:
        cqasm = "version 3; qubit[2] q; X q[0]; wait(2) q[0]; wait(3) q[0]; X q[1]; wait(1) q[1]; X q[0]; X q[1]"
        circuit = Circuit.from_string(cqasm)
        schedule, _ = circuit.export(fmt=ExportFormat.QUANTIFY_SCHEDULER)
        # Check that the last wait is applied correctly
        waited_schedulables = [s for s in schedule.schedulables.values() if "timing_constraints" in s]
        for s in waited_schedulables:
            assert "rel_time" in s["timing_constraints"][0]   
```

rares1609: When it comes to stress-testing the functionality (this is totally unrealistic), maybe a complex circuit that can be tested is:

```cqasm

version 3;
qubit[15] q;
bit[15] b;
init q;

barrier q;
wait(1) q[0]; wait(2) q[1]; wait(3) q[2]; wait(4) q[3]; wait(5) q[4];
wait(1) q[5]; wait(2) q[6]; wait(3) q[7]; wait(4) q[8]; wait(5) q[9];
wait(1) q[10]; wait(2) q[11]; wait(3) q[12]; wait(4) q[13]; wait(5) q[14];

barrier q[0:7];
X q[0]; X q[1]; X q[2]; X q[3]; X q[4]; X q[5]; X q[6]; X q[7];

barrier q[8:14];
CNOT q[8], q[9]; CNOT q[10], q[11]; CNOT q[12], q[13];

wait(2) q[0]; wait(3) q[1]; wait(4) q[2]; wait(5) q[3]; wait(1) q[4];
wait(2) q[5]; wait(3) q[6]; wait(4) q[7]; wait(5) q[8]; wait(1) q[9];
wait(2) q[10]; wait(3) q[11]; wait(4) q[12]; wait(5) q[13]; wait(1) q[14];

barrier q;
CZ q[0], q[14]; CZ q[1], q[13]; CZ q[2], q[12]; CZ q[3], q[11]; CZ q[4], q[10];

barrier q[5:9];
X q[5]; X q[6]; X q[7]; X q[8]; X q[9];

barrier q;
b = measure q;

```
I don't know whether this is an appropriate test to consider since I think nobody would use this for actual experiments. My idea was, if the `quantify_scheduler_exporter` does well on this one, it should be able to handle any arbitrary circuit :). Furthermore, I am not sure either what assertions would be good here. It would be rather unrealistic and maybe redundant for what is important here to stay and compute an `expected_ref_schedulable_indices` by hand for this example.

rares1609: Another good idea I think would be to test barriers on all qubits vs only on some, to see whether references are updated correctly:

```cqasm

qubit[5] q;
bit[5] b;
init q;

barrier q[0:2];
X q[0]; X q[1];

barrier q;
X q[2]; X q[3]; X q[4];

barrier q[1,3];
CNOT q[1], q[3];

barrier q;
b = measure q;
```

For this scenario its also more feasible to compute an `expected_ref_schedulable_indices` by hand I think.

rares1609: An interesting edge case  would be to have a gate applied after a `barrier` on a qubit with `wait`. This test would check whether the functionality is correct in terms of:

- `time_constraint` existing
- `rel_time` matching `wait`
- correct gate schedule `(expected_ref_schedulable_indices` should be computable)

Example circuit:

```cqasm

version 3;
qubit[3] q;
bit[3] b;
init q;

wait(4) q[0];
barrier q[0];
X q[0];

wait(2) q[1];
barrier q[1];
CNOT q[1], q[2];

barrier q;
b = measure q;

```

rares1609 (dismissed):
In my opinion, the implementation itself is good. I was not able to find anything that might critically hinder the functionality, or otherwise improve it. 

Most of my comments focus primarily on possible extra (stress) tests that check whether `quantify_scheduler_exporter` can handle edge cases such as empty circuits, negative `wait` times or overly complex circuits. 

I will mark it as approved for the scenario in which you think the tests are redundant.

tests/passes/exporter/test_quantify_scheduler_exporter.py
elenbaasc: This is a great idea for a test. I will have to check.

elenbaasc: Implemented both, but the first one actually shows incorrect behavior of the processing of the waits; so this requires some changes (WIP).

rares1609 (approved):
This all looks good to me! 💯 I'd say its good to merge.

PR #629, opened 2025-10-15
[CQT-416] Separate control instructions from non unitaries

davidbakker123:


elenbaasc (approved):
Perfect, thanks! 

No further comments :)

PR #630, opened 2025-10-22
[CQT-409] Improve swap placement in routing passes

rares1609:


opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+                source=src,
+                target=dst,
+                heuristic=lambda u, v: calculate_distance(u, v, num_columns, self.distance_metric),
+            )
+        )
+
+    def _plan_swaps(self, ir: IR, graph: nx.Graph, initial_mapping: dict[int, int]) -> list[tuple[int, SWAP]]:
+        # Go through the IR and see where SWAPs need to be placed.
+        swaps_to_insert: list[tuple[int, SWAP]] = []
+        temp_mapping = dict(initial_mapping)
```
elenbaasc: Make sure you don't need to make a copy here explicitly.

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+
+        for statement_idx, statement in enumerate(ir.statements):
             if isinstance(statement, Gate) and len(statement.get_qubit_operands()) == 2:
                 q0, q1 = statement.get_qubit_operands()
-                if not graph.has_edge(q0.index, q1.index):
+                physical_q0 = temp_mapping[q0.index]
+                physical_q1 = temp_mapping[q1.index]
```
elenbaasc: The 'physical' variables are also indices right? _I.e._, not qubits. Make sure to be consistent in the naming of the variables, _e.g._, physical_q0_index.

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
                         msg = f"No routing path available between qubit {q0.index} and qubit {q1.index}"
                         raise NoRoutingPathError(msg) from e
-            statement_index += 1
+
+                    for swap_pair in self._get_swaps_for_path(path):
+                        # Account for earlier planned inserts shifting indices.
+                        insert_position = statement_idx + len(swaps_to_insert)
+                        swaps_to_insert.append((insert_position, SWAP(swap_pair[0], swap_pair[1])))
+                        # Simulate the swap effect for subsequent gates in planning.
+                        self._update_mapping_for_swap(temp_mapping, swap_pair)
```
elenbaasc:
```suggestion
                    for swap_count, swap_pair in enumerate(self._get_swaps_for_path(path)):
                        # Account for earlier planned inserts shifting indices.
                        insert_position = statement_idx + swap_count
                        swaps_to_insert.append((insert_position, SWAP(*swap_pair)))
                        # Simulate the swap effect for subsequent gates in planning.
                        self._update_mapping_for_swap(temp_mapping, swap_pair)
```

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+    def _apply_swaps(
+        self, ir: IR, swaps_to_insert: list[tuple[int, SWAP]], initial_mapping: dict[int, int]
+    ) -> list[Statement]:
+        # Insert planned SWAPs and update qubit indices
+        new_statements: list[Statement] = []
+        swap_index = 0
+        total_swaps = len(swaps_to_insert)
+        final_mapping = dict(initial_mapping)
+
+        for _, statement in enumerate(ir.statements):
```
elenbaasc: Why use `enumerate` if you're not planning to use the index?

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+
+            if isinstance(statement, Instruction):
+                for qubit in statement.get_qubit_operands():
+                    qubit.index = final_mapping[qubit.index]
+
+            new_statements.append(statement)
+
+        while swap_index < total_swaps:
+            new_statements.append(swaps_to_insert[swap_index][1])
+            swap_index += 1
```
elenbaasc: This seems strange to me. We would not expect any SWAPs at the end of a circuit, right?

opensquirrel/passes/router/shortest_path_router.py
```diff
@@ -14,45 +13,104 @@ def __init__(self, connectivity: dict[str, list[int]], **kwargs: Any) -> None:
...
-            # Update subsequent statements to reflect the swap
-            for statement in ir.statements[statement_index + 1 :]:
-                if isinstance(statement, Instruction):
-                    for qubit in statement.get_qubit_operands():
-                        if qubit.index == start_qubit_index:
-                            qubit.index = end_qubit_index
-                        elif qubit.index == end_qubit_index:
-                            qubit.index = start_qubit_index
-            statement_index += 1
+    def _get_swaps_for_path(self, shortest_path: list[int]) -> list[tuple[int, int]]:
```
elenbaasc: Many of these helper functions are not particular to the type of router (i.e. A* or shortest-path), so it would be better to create a utility/common module in the router folder with these functionalities. Thereby also decreasing code duplication.

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+        total_swaps = len(swaps_to_insert)
+        final_mapping = dict(initial_mapping)
+
+        for _, statement in enumerate(ir.statements):
+            while swap_index < total_swaps and swaps_to_insert[swap_index][0] == len(new_statements):
+                swap_gate = swaps_to_insert[swap_index][1]
+                new_statements.append(swap_gate)
+
+                swap_qubits = swap_gate.get_qubit_operands()
+                swap_pair = (swap_qubits[0].index, swap_qubits[1].index)
```
elenbaasc:
```suggestion
                swap_pair = tuple([qubit.index for qubit in swap_gate.get_qubit_operands()])
```

opensquirrel/passes/router/astar_router.py
```diff
@@ -21,54 +20,120 @@ def __init__(
...
+
+    def route(self, ir: IR) -> IR:
+        """
+        Routes the circuit by inserting SWAP gates (via A*) to satisfy hardware connectivity.
+        """
+        graph = self._build_graph()
+        initial_mapping = self._init_qubit_mapping(ir)
+        swaps_to_insert = self._plan_swaps(ir, graph, initial_mapping)
+        ir.statements = self._apply_swaps(ir, swaps_to_insert, initial_mapping)
         return ir
```
elenbaasc: Consider using object variables, i.e. `self._graph`, `self._initial_mapping`, `self._ir` and update those, instead of passing them and returning them. (passing and returning is probably necessary when moving many of these general routing functionalities to a utility/common module.

Also, the `ir` is returned here; is that necessary?

elenbaasc (commented):
Nice implementation :), thanks for the effort!

Mostly comments about rewritting certain functionalities and moving common code to a separate module.

opensquirrel/passes/router/astar_router.py
rares1609: I am not entirely sure how I should go about this one. Maybe better to come back to this after I commit the new version with a `common` file containing the helper functions

elenbaasc (approved):

eendebakpt:
@rares1609 @elenbaasc This PR added `uv.lock` to the repo. I think this should be reverted (and `uv.lock` added to the `.gitignore` file).

PR #653, opened 2025-12-23
[CQT-437] Separate arguments into parameters (or arguments) and operands

davidbakker123:


opensquirrel/ir/two_qubit_gate.py
```diff
@@ -63,10 +63,15 @@ def accept(self, visitor: IRVisitor) -> Any:
 
     @property
     def arguments(self) -> tuple[Expression, ...]:
+        return ()
```
elenbaasc: Why does this return an empty tuple? Could this be a `pass`? If so, then it inherits it from `Instruction`.

opensquirrel/ir/statement.py
```diff
@@ -40,8 +40,18 @@ def __init__(self, name: str) -> None:
...
         pass
 
+    @property
+    @abstractmethod
+    def qubit_operands(self) -> tuple[Qubit, ...]:
+        pass
+
+    @property
+    def qubit_indices(self) -> list[int]:
+        return [qubit.index for qubit in self.qubit_operands]
```
elenbaasc, line 50: Probably nice to have indeed :)

opensquirrel/ir/unitary.py
```diff
@@ -43,7 +39,7 @@ def __eq__(self, other: object) -> bool:
 
 
 def compare_gates(g1: Gate, g2: Gate) -> bool:
-    union_mapping = [q.index for q in list(set(g1.get_qubit_operands()) | set(g2.get_qubit_operands()))]
+    union_mapping = [q.index for q in list(set(g1.qubit_operands) | set(g2.qubit_operands))]
```
elenbaasc: Could you use the `qubit_indices` here?

opensquirrel/passes/decomposer/general_decomposer.py
```diff
@@ -20,15 +20,15 @@ def decompose(self, gate: Gate) -> list[Gate]:
...
 def check_gate_replacement(gate: Gate, replacement_gates: Iterable[Gate]) -> None:
-    gate_qubit_indices = [q.index for q in gate.get_qubit_operands()]
+    gate_qubit_indices = gate.qubit_indices
     replacement_gates_qubit_indices = set()
     replaced_matrix = get_circuit_matrix(get_reindexed_circuit([gate], gate_qubit_indices))
 
     if is_identity_matrix_up_to_a_global_phase(replaced_matrix):
         return
 
     for g in replacement_gates:
```
elenbaasc:
```suggestion
    for replacement_gate in replacement_gates:
```

opensquirrel/passes/decomposer/general_decomposer.py
```diff
@@ -20,15 +20,15 @@ def decompose(self, gate: Gate) -> list[Gate]:
...
-        replacement_gates_qubit_indices.update([q.index for q in g.get_qubit_operands()])
+        replacement_gates_qubit_indices.update(g.qubit_indices)
```
elenbaasc:
```suggestion
        replacement_gates_qubit_indices.update(replacement_gate.qubit_indices)
```

opensquirrel/passes/exporter/cqasmv1_exporter/cqasmv1_exporter.py
```diff
@@ -107,14 +107,11 @@ def visit_two_qubit_gate(self, gate: TwoQubitGate) -> Any:
...
 
-        if len(gate.arguments) == 2:
-            self.output += f"{gate.name.lower()} {qubit_operand_0}, {qubit_operand_1}\n"
-        elif len(gate.arguments) == 3:
-            _, _, arg = gate.arguments
-            argument = arg.accept(self)
-            self.output += f"{gate.name.lower()}({argument}) {qubit_operand_0}, {qubit_operand_1}\n"
+        if gate.arguments:
+            arguments = ", ".join(arg.accept(self) for arg in gate.arguments)
+            self.output += f"{gate.name.lower()}({arguments}) {qubit_operand_0}, {qubit_operand_1}\n"
```
elenbaasc, line 112: Yes, this is a lot nicer now :)

elenbaasc (commented):
Nice changes, thanks!

Only minor comments.

opensquirrel/ir/two_qubit_gate.py
davidbakker123: `arguments` is an abstract method of `Instruction`. So `TwoQubitGate` has to implement it, otherwise it would have to be an abstract class, which is not something that we want, I think. Or you implement `arguments` in the `Gate` class.

opensquirrel/ir/unitary.py
davidbakker123: Yes!

opensquirrel/ir/two_qubit_gate.py
elenbaasc: Thanks for clarifying. Both `arguments` and `bit_operands` can be moved to the `Gate` class (and both return an empty tuple).

elenbaasc (commented):
Just one minor change and it's ready to be merged!

opensquirrel/ir/two_qubit_gate.py
davidbakker123: Done. Also removed `arguments` and `bit_operands` from `SingleQubitGate`.

elenbaasc (approved):
All good!

PR #657, opened 2026-01-09
[CQT-413] Refactor register manager to be agnostic of libQASM/cQASM

elenbaasc:


opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
+class Register:
+    """Register manages a (virtual) register."""
 
-    def __repr__(self) -> str:
-        return "[{}{}]".format(self.first, "" if self.size == 1 else f"..{self.first + self.size - 1}")
+    def __init__(
+        self,
+        size: int,
+        name: str,
+        virtual_index_0: int = 0,
```
davidbakker123: Maybe rename `virtual_index_0` to `virtual_zero_index`.

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
-        return self.index_to_variable_name[index]
-
-    def get_range(self, variable_name: str) -> Range:
-        """Get the Range for a `variable_name`."""
-        return self.variable_name_to_range[variable_name]
-
-    def get_index(self, variable_name: str, sub_index: int) -> int:
-        """Get the Index for a given `subIndex` of a `variable_name`."""
-        return self.variable_name_to_range[variable_name].first + sub_index
+        return self._size == other.size and self._name == other.name and self._virtual_index_0 == other.virtual_index_0
```
davidbakker123: Now it doesn't matter if you use the protected attribute but if you ever have a more elaborate implementation of the property you might not compare the same things.

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
-        index_to_variable_name: dict[int, str] | None = None,
-    ) -> None:
-        self.register_size: int = register_size
-        self.variable_name_to_range: dict[str, Range] = variable_name_to_range or {}
-        self.index_to_variable_name: dict[int, str] = index_to_variable_name or {}
+    def __getitem__(self, index: int) -> Any:
+        if index > self._size:
+            msg = f"Index {index} is out of range"
+            raise IndexError(msg)
+        return self._virtual_index_0 + index
```
davidbakker123: Do you also want to check if `index < self._virtual_index_0`?

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
 
-from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from collections import OrderedDict
 from typing import Any
 
-import cqasm.v3x as cqasm
-from typing_extensions import Self
+QUBIT_REGISTER_NAME = "q"
+BIT_REGISTER_NAME = "b"
```
davidbakker123: Add `DEFAULT_` as prefix?

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
+        for register in registers:
+            register.virtual_index_0 = virtual_index
+            virtual_index += register.size
+        return register_cls(virtual_index)
+
+    def add_qubit_register(self, qubit_register: QubitRegister) -> None:
+        self._qubit_registry[qubit_register.name] = qubit_register
+
+    def add_bit_register(self, bit_register: BitRegister) -> None:
+        self._bit_registry[bit_register.name] = bit_register
```
davidbakker123, line 107: Check if `(qu)bit_register.name` already exists in the registry.

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,134 @@
...
 
-    def get_qubit_index(self, variable_name: str, sub_index: int) -> int:
-        return self.qubit_register.get_index(variable_name, sub_index)
+    def get_qubit_register(self, qubit_register_name: str) -> QubitRegister:
+        return self._qubit_registry[qubit_register_name]  # type: ignore[return-value]
 
-    def get_bit_index(self, variable_name: str, sub_index: int) -> int:
-        return self.bit_register.get_index(variable_name, sub_index)
+    def get_bit_register(self, bit_register_name: str) -> BitRegister:
+        return self._bit_registry[bit_register_name]  # type: ignore[return-value]
```
davidbakker123, line 129: Why do you ignore here?

opensquirrel/reader/libqasm_parser.py
```diff
@@ -244,4 +263,7 @@ def circuit_from_string(self, s: str) -> Circuit:
                     self.ir.add_statement(instruction_generator(*args))
                 expanded_args = []
 
+        if not self.register_manager:
+            msg = "parsing error: no registers found"
+            raise OSError(msg)
```
davidbakker123: Why not do this check right after creating `self.register_manager` on line 235?

opensquirrel/reader/libqasm_parser.py
```diff
@@ -213,7 +232,7 @@ def circuit_from_string(self, s: str) -> Circuit:
         ast = analysis_result
 
         # Create RegisterManager
-        self.register_manager = RegisterManager.from_ast(ast)
+        self.register_manager: RegisterManager = self._create_register_manager(ast)
```
davidbakker123: Do you actually need this type here? Weird...

opensquirrel/reader/libqasm_parser.py
```diff
@@ -205,6 +208,22 @@ def _get_non_gate_instruction_generator(
...
         return lambda *args: default_non_unitary_set[instruction.name](*args)
 
+    @staticmethod
+    def _get_registry(
+        ast: Any,
+        register_cls: type[QubitRegister | BitRegister],
+        type_check: Callable[[Any], bool],
+    ) -> Registry:
+        registry = OrderedDict()
+        for variable in [v for v in ast.variables if type_check(v.typ)]:
```
davidbakker123: If you modify `_is_(qu)bit_type` to receive `v.typ` instead, you can change this line to `for variable in filter(type_check, ast.variables):`

opensquirrel/passes/mapper/check_mapper.py
```diff
@@ -36,7 +43,9 @@ def check_mapper(mapper: Mapper) -> None:
     """
     assert isinstance(mapper, Mapper)
 
-    register_manager = RegisterManager(QubitRegister(10), BitRegister(10))
+    register_manager = RegisterManager(
+        OrderedDict({QUBIT_REGISTER_NAME: QubitRegister(10)}), OrderedDict({BIT_REGISTER_NAME: BitRegister(10)})
```
davidbakker123: Where does this 10 come from?

opensquirrel/passes/exporter/cqasmv1_exporter.py
davidbakker123, line 1: I actually wanted to ask why `cqasmv1_exporter.py` was in a separate folder and `quantify_scheduler_exporter.py` not. So good change! xD

davidbakker123 (commented):
Nice refactor! Just a few comments :)

opensquirrel/register_manager.py
elenbaasc: Very good point, thanks for noting! I changed it to using the public variables.

opensquirrel/register_manager.py
elenbaasc: The relation of `index` to `self._virtual_index_zero` is not relevant. The latter is only used to map the logical index to the virtual index.

But, now that you mentioned this, I have adjusted the `__getitem__` to handle negative input indices properly.

opensquirrel/passes/mapper/check_mapper.py
elenbaasc: This function checks if new implementations of a Mapper are compatible with the `Circuit.map` functionality. So the number 10 is somewhat arbitrary, just to a have registers that can be remapped.

opensquirrel/reader/libqasm_parser.py
elenbaasc: `mypy` required it because the `self.register_manager` was not initialized in the init, but now there is the check `if not self.register_manager`, it is apparently no longer required.

opensquirrel/register_manager.py
davidbakker123: I will check this out in [CQT-436]

opensquirrel/register_manager.py
```diff
@@ -1,187 +1,143 @@
...
-    ) -> None:
-        self.register_size: int = register_size
-        self.variable_name_to_range: dict[str, Range] = variable_name_to_range or {}
-        self.index_to_variable_name: dict[int, str] = index_to_variable_name or {}
+    def __getitem__(self, index: int) -> Any:
+        if abs(index) > self._size:
+            msg = f"Index {index} is out of range"
+            raise IndexError(msg)
+        size = self._size if index < 0 else 0
+        return self._virtual_zero_index + index + size
```
davidbakker123, line 44: Nice!

davidbakker123 (approved):
Ready to go!

PR #658, opened 2026-01-13
[CQT-417] Circuit builder accepts multiple registers

elenbaasc:


opensquirrel/circuit_builder.py
```diff
@@ -62,6 +77,12 @@ def __getattr__(self, attr: str) -> Any:
         # Default behaviour
         return self.__getattribute__(attr)
 
+    def add_register(self, register: Register) -> None:
+        if isinstance(register, QubitRegister):
+            self.register_manager.add_qubit_register(register)
+        if isinstance(register, BitRegister):
+            self.register_manager.add_bit_register(register)
```
davidbakker123: Maybe move this to the `RegisterManager`, so add a method `add_register` to it.

tests/test_circuit_builder.py
```diff
@@ -102,3 +129,206 @@ def test_int_qubit_parsing(self) -> None:
...
+
+        builder.add_register(q1)
+        builder.add_register(q2)
+        builder.add_register(q3)
+
+        q1_index = 3
+        q2_index = 5
+        q3_index = 9
+
+        assert q1[q1_index] == q1[q1_index]
```
rares1609: I don't really understand the purpose of this assert statement.

tests/test_circuit_builder.py
rares1609, line 1: The tests here look good! The only thought I would have to test it a bit further, is to have an example in which one defines a circuit, adds a couple of (qu)bit registers, then some instructions on said registers, and afterwards another set of registers + maybe instructions, to then see if the resulting circuit is as expected.

opensquirrel/reader/libqasm_parser.py
```diff
@@ -23,7 +23,12 @@
     Qubit,
     Statement,
 )
-from opensquirrel.register_manager import BitRegister, QubitRegister, RegisterManager, Registry
+from opensquirrel.register_manager import (
+    BitRegister,
+    QubitRegister,
+    RegisterManager,
+    Registry,
```
rares1609, line 32: This more of a question for my own knowledge, but what is the difference between between a register and a registry? I see in `register_manager.py` that a registry is an `OrderedDict`, but I am not entirely sure what is the difference say, between a `bit_registry` and a `bit_register`. And in the case there is not a particular difference, wouldn't it be better maybe to stick to a single naming?

rares1609 (changes requested):
The issue looks good to me. The `CircuitBuilder` now accepts multiple (qu)bit registers which can be inserted in-between instructions. I think there are two (minor) points that still need to be addressed before its merged:

- The comment concerning the assert statement, I am not certain whether it is redundant or done on purpose.
- An entry needs to be added to the Changelog.

opensquirrel/reader/libqasm_parser.py
```diff
@@ -216,12 +221,12 @@ def _get_registry(
...
         for variable in filter(type_check, ast.variables):
             registry[variable.name] = register_cls(variable.typ.size, variable.name)
-        return registry
+        return registry or OrderedDict({register_cls.default_name: register_cls(0)})
 
     def _create_register_manager(self, ast: Any) -> RegisterManager:
         qubit_registry = self._get_registry(ast, QubitRegister, LibQasmParser._is_qubit_type)
         bit_registry = self._get_registry(ast, BitRegister, LibQasmParser._is_bit_type)
-        return RegisterManager(qubit_registry, bit_registry)
+        return RegisterManager(qubit_registry, bit_registry)  # type: ignore [arg-type]
```
davidbakker123: You can use `@overload` to fix this typing issue.

davidbakker123 (commented):
I was half way through the review, but rares finished. So I will leave it here :)

opensquirrel/reader/libqasm_parser.py
davidbakker123:
```
    @overload
    @staticmethod
    def _get_registry(
        ast: Any,
        register_cls: type[QubitRegister],
        type_check: Callable[[Any], bool],
    ) -> QubitRegistry: ... 

    @overload  
    @staticmethod
    def _get_registry(
        ast: Any,
        register_cls: type[BitRegister],
        type_check: Callable[[Any], bool],
    ) -> BitRegistry: ...
        

    @staticmethod
    def _get_registry(
        ast: Any,
        register_cls: type[QubitRegister | BitRegister],
        type_check: Callable[[Any], bool],
    ) -> Registry:
        registry = OrderedDict()
        for variable in filter(type_check, ast.variables):
            registry[variable.name] = register_cls(variable.typ.size, variable.name)
        return registry or OrderedDict({register_cls.default_name: register_cls(0)})

    def _create_register_manager(self, ast: Any) -> RegisterManager:
        qubit_registry = self._get_registry(ast, QubitRegister, LibQasmParser._is_qubit_type)
        bit_registry = self._get_registry(ast, BitRegister, LibQasmParser._is_bit_type)
        return RegisterManager(qubit_registry, bit_registry)  
 ```

tests/test_circuit_builder.py
elenbaasc: Good catch! It should be `q1[q1_index] == q1_index`, otherwise it is completely trivial.

tests/test_circuit_builder.py
elenbaasc: I have added this scenario to the `test_multiple_registers` test.

opensquirrel/reader/libqasm_parser.py
elenbaasc: A __Register__ has a `name`, `size`, and `virtual_zero_index`; it represents a _list_ of (qu)bits with specific defined behaviour to _manipulate_ that list. A __Registry__ is a _dictionary_ of registers; it's an (ordered) record of registers that are identified by their name. The `RegisterManager` needs to record all the registers that are in the circuit and generate the virtual register from all these registers. E.g., a (qu)bit registry holds all (qu)bit registers that are part of the circuit, and the `RegisterManager` generates the virtual (qu)bit register from the (qu)bit registry of (qu)bit registers.

rares1609: Thank you for the explanation! Understood now 👍

opensquirrel/reader/libqasm_parser.py
elenbaasc: Thanks for the tip! I have used the implementation you provided 😄

rares1609 (approved):
This looks good to me! 👍 The `CircuitBuilder` can now accept multiple (qu)bit registers, and these can be inserted at arbitrary points in the design of a circuit. Furthermore, the changelog is also updated with this added feature. I say its good to be merged.

PR #659, opened 2026-01-16
[CQT-414] mip mapper maps circuits that do not need to be remapped

rares1609:


opensquirrel/passes/mapper/mapping.py
```diff
@@ -22,8 +22,19 @@ def __init__(self, physical_qubit_register: list[int]) -> None:
...
+
+        if set(self.data.keys()) != set(other.data.keys()):
+            return False
+
+        for key in self.data:
+            if self.data[key] != other.data[key]:
+                return False
+
+        if self.data != other.data:
+            return False
```
elenbaasc: This is redundant if you follow it with `return self.data == other.data`

```suggestion
```

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -1,7 +1,8 @@
-# OpenQL MIP-Like Mapper
+# OpenQL MIP-Like Mapper. This mapper pass takes inspiration from:
```
elenbaasc: Is based on? 

"_... takes inspiration from_" is a bit vague.

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
 
         return list(distance)
 
-    def _get_cost(
-        self, ir: IR, distance: list[list[int]], num_virtual_qubits: int, num_physical_qubits: int
-    ) -> list[list[int]]:
+    def _get_reference_counter(self, ir: IR, num_virtual_qubits: int) -> list[list[int]]:
         reference_counter = [[0 for _ in range(num_virtual_qubits)] for _ in range(num_virtual_qubits)]
```
elenbaasc, line 131: Why does the following replacement not work in this case? (it seems to do the same thing, but when I run the tests they fail..., which is a bit dodgy.)
```suggestion
        reference_counter = [[0] * num_virtual_qubits] * num_virtual_qubits
```

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
 
         return list(distance)
 
-    def _get_cost(
-        self, ir: IR, distance: list[list[int]], num_virtual_qubits: int, num_physical_qubits: int
-    ) -> list[list[int]]:
+    def _get_reference_counter(self, ir: IR, num_virtual_qubits: int) -> list[list[int]]:
```
elenbaasc: This is a static method:
```suggestion
    @staticmethod
    def _get_reference_counter(ir: IR, num_virtual_qubits: int) -> list[list[int]]:
```

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
         reference_counter = [[0 for _ in range(num_virtual_qubits)] for _ in range(num_virtual_qubits)]
         for statement in getattr(ir, "statements", []):
```
elenbaasc: The following should be fine, right?
```suggestion
        for statement in ir.statements:
```

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+            w[i][k]: continuous, cost contribution for mapping i to k (num_virtual * num_physical)
+
+        Objective:
+            min sum(w[i][k]) + epsilon * sum_{i,k: i!=k} x[i][k]
+            (small epsilon penalty for non-identity mappings as tiebreaker)
+
+        Constraints:
+            1. Each virtual qubit assigned to exactly one physical qubit: sum_k x[i][k] == 1
+            2. Each physical qubit assigned to at most one virtual qubit: sum_i x[i][k] <= 1
+            3. Linearization: costmax[i][k] * x[i][k] + sum_{j,l} refcount[i][j]*distance[k][l]*x[j][l] - w[i][k] <= costmax[i][k]
```
elenbaasc: I guess these are part of the main description.

Recognized subheadings are:

- Args
- Returns
- Yields
- Raises
- Attributes
- Example / Examples
- Note
- Notes
- See Also
- Todo
- Warning / Warnings

We shouldn't introduce new ones (that might be a bit confusing)

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+        """  # noqa: E501
+        num_x_vars = num_virtual_qubits * num_physical_qubits
+        num_w_vars = num_virtual_qubits * num_physical_qubits
+        num_vars = num_x_vars + num_w_vars
+
+        costmax = self._compute_costmax(reference_counter, distance, num_virtual_qubits, num_physical_qubits)
```
elenbaasc: `max_cost`  and `_compute_max_cost`?

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+                costmax[i][k] = sum(
                     reference_counter[i][j] * distance[k][l]
                     for j in range(num_virtual_qubits)
                     for l in range(num_physical_qubits)  # noqa: E741
                 )
-        return cost
+        return costmax
+
+    @staticmethod
+    def _get_cost(num_virtual_qubits: int, num_physical_qubits: int, num_x_vars: int, num_w_vars: int) -> np.ndarray:
```
elenbaasc: This returns an numpy array, but the only place where it is used it is directly converted to a list... also no operations are done on the entire array, so maybe a simple list would be more fitting...

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+        epsilon = 1e-6
```
elenbaasc: This epsilon needs to be define globally and as a constant (or maybe a class variable, so that the user can adjust it).

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+        num_vars: int,
+        num_x_vars: int,
+    ) -> list[LinearConstraint]:
         eq_a = np.zeros((num_virtual_qubits, num_vars))
-        for q_i in range(num_virtual_qubits):
-            for q_k in range(num_physical_qubits):
-                eq_a[q_i, q_i * num_physical_qubits + q_k] = 1
+        for i in range(num_virtual_qubits):
+            for k in range(num_physical_qubits):
+                eq_a[i, i * num_physical_qubits + k] = 1
```
elenbaasc: Why remove the `q_`, does that make more sense?

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -86,44 +91,134 @@ def _get_distance(connectivity: dict[str, list[int]]) -> list[list[int]]:
...
+                constraint_idx += 1
+
+        return [
+            LinearConstraint(eq_a, eq_b, eq_b),
+            LinearConstraint(ub_a, -np.inf, ub_b),
+            LinearConstraint(lin_a, -np.inf, lin_b),
+        ]
+
+    @staticmethod
+    def _get_integrality_and_bounds(num_x_vars: int, num_w_vars: int) -> tuple[np.ndarray, Bounds]:
```
elenbaasc: Also here, the `integrality` is returned as a numpy array of booleans, but where it is used it expects a list of integers. Why not define it as a list of integers from the start?

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -60,8 +61,12 @@ def map(self, ir: IR, qubit_register_size: int) -> Mapping:
             raise RuntimeError(error_message)
 
         distance = self._get_distance(self.connectivity)
-        cost = self._get_cost(ir, distance, qubit_register_size, num_physical_qubits)
-        constraints, integrality, bounds = self._get_constraints(qubit_register_size, num_physical_qubits)
+        reference_counter = self._get_reference_counter(ir, qubit_register_size)
+
+        cost, constraints, integrality, bounds = self._get_linearized_formulation(
+            reference_counter, distance, qubit_register_size, num_physical_qubits
```
elenbaasc: The `qubit_register_size` and `num_physical_qubits` can probably be defined as object attributes, i.e. `self.qubit_register_size` and `self.num_physical_qubits`, then they don't need to be passed around.

Same goes for `num_virtual_qubits * num_physical_qubits`, which is used in various places.

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -151,10 +244,12 @@ def _solve_and_extract_mapping(
...
+
         mapping = []
-        for q_i in range(num_virtual_qubits):
-            q_k = int(np.argmax(x_sol[q_i]))
-            mapping.append(q_k)
+        for i in range(num_virtual_qubits):
+            k = int(np.argmax(x_sol[i]))
+            mapping.append(k)
 
         return mapping
```
elenbaasc:
```suggestion
        return list(map(lambda x: int(np.argmax(x)), x_sol))
```

tests/passes/mapper/test_mip_mapper.py
```diff
@@ -106,7 +132,7 @@ def test_timeout(mapper3: MIPMapper, circuit2: Circuit) -> None:
 def test_map_method(mapper1: MIPMapper, circuit1: Circuit) -> None:
     initial_circuit = str(circuit1)
     circuit1.map(mapper=mapper1)
-    assert str(circuit1) != initial_circuit
+    assert str(circuit1) == initial_circuit
```
elenbaasc: I would prefer a different mapping to be the result of a `test_map_method`. Nevertheless, there should also be a test where the map method is applied, and as expected, the mapping remains the same (just like you did in the `test_controlled_gates`).

tests/passes/mapper/test_simple_mappers.py
```diff
@@ -27,7 +27,7 @@ def test_compliance(self, mapper: IdentityMapper) -> None:
 
     def test_get_mapping(self, mapper: IdentityMapper, circuit: Circuit) -> None:
         mapping = mapper.map(circuit.ir, circuit.qubit_register_size)
-        assert mapping == Mapping([0, 1, 2])
+        assert mapping.items() == Mapping([0, 1, 2]).items()
```
elenbaasc: Is this still needed, now that you updated the `Mapping.__eq__` method? Maybe this condition `self.data == other.data` should not be part of that method.

elenbaasc (commented):
Thanks for the effort! It seems it was a bit more work that I had anticipated :)

Mostly minor comments. Mainly in regards to the implementation of the map method where you default to using numpy arrays, when simple objects would suffice or are expected in the end anyway, e.g. creating a numpy array and at the end converting it to a list.

opensquirrel/passes/mapper/mip_mapper.py
rares1609: It absolutely is. I don't know why, I used `getattr` at some point in one of my first issues for accessing the instructions, and it just stuck to me 😆

opensquirrel/passes/mapper/mip_mapper.py
rares1609: Yes precisely, I wrote this docstring to properly explain the mathematical formulation of this, as it is also done in the CPP docstring that I based this implementation on.

But indeed, I should re-write this to only include proper subheadings.

opensquirrel/passes/mapper/mip_mapper.py
rares1609: For now I've defined it as a class variable.

opensquirrel/passes/mapper/mip_mapper.py
rares1609: I did this solely because of habit (i.e. using `i` when iterating over something :)). I will re-name to `q_i` and `q_k`, it doesn't make any extra sense to rename them to just `i` and `k`.

opensquirrel/passes/mapper/mip_mapper.py
rares1609: Thanks for this suggestion! Looks much nicer indeed.

tests/passes/mapper/test_mip_mapper.py
rares1609: In that case (of course, correct me if I am wrong), should we remove the `test_map_method` altogether? I am suggesting this since in `test_controlled_gates` we test the functionality of `circuit.map()` (for the `MipMapper`) rather thoroughly, with both circuits initialised from a string, and with the `CircuitBuilder`, for scenarios in which circuit remapping is needed and in which it is not.

tests/passes/mapper/test_simple_mappers.py
rares1609: Indeed this is no longer needed, so i removed the `.items()` from the assertions, in both `test_simple_mappers.py` and `test_mip_mapper.py`. Regarding the condition in `Mapping.__eq__` what should be done about it? Should this be part of a different issue?

tests/passes/mapper/test_mip_mapper.py
elenbaasc: Yes, it's fine to remove `test_map_method`.

tests/passes/mapper/test_simple_mappers.py
elenbaasc: Good :)

Removing the following lines from the `Mapping.__eq__`:
```
if self.data != other.data:
            return False
```
should do the trick. (Fingers crossed that the tests succeed...)

rares1609: Thankfully they work 😆

opensquirrel/passes/mapper/mip_mapper.py
rares1609: To be fully honest with you, I am not exactly sure why the suggested replacement does not work. **I think** the suggested version creates shallow copies of the list object in memory, and something goes on there?

opensquirrel/passes/mapper/mip_mapper.py
elenbaasc: This is still not a static method.

opensquirrel/passes/mapper/mip_mapper.py
```diff
@@ -50,80 +58,169 @@ def map(self, ir: IR, qubit_register_size: int) -> Mapping:
...
             if isinstance(statement, TwoQubitGate):
-                for q_0, q_1 in itertools.pairwise(statement.qubit_operands):
+                qubit_operands = statement.qubit_operands
+                if len(qubit_operands) == 2:
+                    q_0, q_1 = qubit_operands[0], qubit_operands[1]
                     reference_counter[q_0.index][q_1.index] += 1
                     reference_counter[q_1.index][q_0.index] += 1
+        return reference_counter
+
+    def _get_linearized_formulation(
```
elenbaasc, line 141: Shouldn't these docstring be part of the `map` method (as it is the only public method of this class)? It might even be the docstring of the class... docstrings in private classes are not published in the user docs, and are only there for the developers.

tests/passes/mapper/test_simple_mappers.py
elenbaasc: I see that you removed:
```python
for key in self.data:
    if self.data[key] != other.data[key]:
        return False
```
but that the following is still there:
```python
return self.data == other.data
```
The first one should be there, _i.e._, you are checking if all _values_ match. The second one should __not__ be there, because you don't want to say that the _object_ `self.data` must be equal to the _object_ `other.data`, for one mapping to be equal to another.

elenbaasc (commented):
Thanks for the effort! Almost there :)

Some comments not fully resolved yet and a small comments about the placement of the docstring.

opensquirrel/passes/mapper/mip_mapper.py
rares1609: Sorry about that, could have sworn I made it static before commiting 😢

tests/passes/mapper/test_simple_mappers.py
rares1609: I replaced the statement checking if all values match with `return all(self.data[key] == other.data[key] for key in self.data)`.

opensquirrel/passes/mapper/mapping.py
```diff
@@ -22,9 +22,13 @@ def __init__(self, physical_qubit_register: list[int]) -> None:
             raise ValueError(msg)
 
     def __eq__(self, other: Any) -> bool:
-        if not isinstance(other, Mapping):
```
elenbaasc, line 25: Why remove this check?

pyproject.toml
```diff
@@ -43,6 +43,7 @@ dev = [
     "sympy>=1.14.0,<2",
     "tox>=4.30.2,<5",
     "ty>=0.0.10",
+    "cachetools<=6.2.4",
```
elenbaasc: Is this still necessary given your fix?

tests/passes/mapper/test_mip_mapper.py
```diff
@@ -80,15 +82,39 @@ def circuit3() -> Circuit:
...
+    assert computed_mapping == expected_mapping
+
+
+def test_mip_mapper_remaps_when_needed(mapper2: MIPMapper, circuit2: Circuit) -> None:
+    if sys.platform.startswith("linux") or sys.platform == "win32":
+        expected_mapping = Mapping([2, 1, 3, 0, 4, 5, 6])
+    elif sys.platform == "darwin":  # pragma: no cover
+        expected_mapping = Mapping([3, 4, 2, 0, 1, 5, 6])
+    else:  # pragma: no cover
+        pytest.skip(f"Unknown platform: {sys.platform}")
```
elenbaasc, line 115: Do I understand correctly that, based on the platform, the mapping will be different (but that for all platforms a different mapping is created, i.e., not the identity mapping is returned?) This is because of the packages used by the MIP mapper right?

And just to double-check: for all platforms, if the mapping is sufficient, the MIP mapper does not re-map, right?

elenbaasc (commented):
Very nice to see that you found a fix to the issue!!

I still have some minor comments/checks. You're very close to merging 😄

opensquirrel/passes/mapper/mapping.py
rares1609: Quite wrong of me indeed to remove it, as it is an important check. I've added it back.

pyproject.toml
rares1609: Seems to work without it.

tests/passes/mapper/test_mip_mapper.py
rares1609: For the first question, yes, from what I read and understand, the main package which now demands this check is `scipy`'s `milp`. The reason for which a different mapping is computed for this case on `macos` systems is that milp uses [HiGHS](https://highs.dev/) for linear optimisation. From what I understand now, depending on the system, a different optimal/high-quality mapping may be chosen because of floating-point differences, optimisation strategies or tie-breaking rules. But I can't say, and I also don't think its a certainty, that for every single example of a circuit that needs to be remapped given a target architecture, the mapping obtained on `macos` systems will be different compared to mappings computed in `windows` and `linux` every time. I can think of different examples to test this, but I think there should be plenty of scenarios in which the computed mapping is consistent across operating systems.

For the second question, once again, yes! When the identity mapping is the optimal mapping, the mapper returns the identity mapping on all platforms.

elenbaasc (approved):

PR #668, opened 2026-02-10
[CQT-443] Make target gate of a controlled gate semantic a BSR rotation semantic

davidbakker123:


opensquirrel/ir/default_gates/two_qubit_gates.py
```diff
@@ -57,7 +56,10 @@ def __init__(self, control_qubit: QubitLike, target_qubit: QubitLike) -> None:
 class CZ(TwoQubitGate):
     def __init__(self, control_qubit: QubitLike, target_qubit: QubitLike) -> None:
         super().__init__(
-            qubit0=control_qubit, qubit1=target_qubit, gate_semantic=ControlledGateSemantic(Z(target_qubit)), name="CZ"
+            qubit0=control_qubit,
+            qubit1=target_qubit,
+            gate_semantic=ControlledGateSemantic(BsrNoParams(axis=(0, 0, 1), angle=pi, phase=pi / 2)),
```
elenbaasc, line 61: Would the following work?
```suggestion
            gate_semantic=ControlledGateSemantic(Z.bsr),
```
Not that it's necessary, but might make it a _bit_ more readable and less error prone. (Sorry to mention it now, after you already made all the changes...)

opensquirrel/utils/matrix_expander.py
```diff
@@ -136,14 +136,32 @@ def _controlled_gate(self, gate: TwoQubitGate) -> NDArray[np.complex128]:
             msg = f"gate {gate!r} does not have a controlled gate semantic"
             raise ValueError(msg)
 
-        control_qubit = gate.qubit0
-        target_gate = gate.controlled.target_gate
+        qubit_operands = list(reversed(gate.qubit_operands))
```
elenbaasc, line 139: Why reverse it? Can line 141 be changed to `if any(q.index >= self.qubit_register_size for q in gate.qubit_operands):` without first explicitly declaring `qubit_operands`?

elenbaasc (commented):
Looks good!

Just two small comments remain ✌️

opensquirrel/ir/default_gates/two_qubit_gates.py
davidbakker123: It would not work like this, because it only has the `bsr` attribute after initialization of the class. So you would need to do this: `gate_semantic=ControlledGateSemantic(Z(0).bsr)`

elenbaasc: Understood!

opensquirrel/utils/matrix_expander.py
davidbakker123: Yes, this can be removed. It was left over from a older function. I didn't know if it would break something, so I just left it there. :)

elenbaasc (approved):
Looks good!

Can be merged once the pipeline is all green 😄

PR #672, opened 2026-03-03
[CQT-429] Add mapping attribute to circuit object in opensquirrel

rares1609:


opensquirrel/circuit.py
```diff
@@ -134,6 +137,9 @@ def interaction_graph(self) -> InteractionGraph:
                     graph[edge] = graph.get(edge, 0) + 1
         return graph
 
+    def _set_identity_mapping(self) -> Mapping:
+        return IdentityMapper().map(self, self.qubit_register_size)
+
```
davidbakker123: This doesn't have to be a method. I don't see the issue with putting this directly into the `__init__`

opensquirrel/circuit.py
```diff
@@ -167,6 +173,8 @@ def map(self, mapper: Mapper) -> None:
 
         remap_ir(self, mapping)
 
+        self.mapping = mapping
```
davidbakker123: Can you not assign it immediately on line 172, or will this break something?

davidbakker123 (commented):
Two final things,
1. The ticket says the circuit should have an empty mapping. I guess that is the same as the identity mapping?
2. You might want to consider using this feature to update the tests where this was an "issue" (like mentioned in the ticket). Maybe it is too much work and you can make a new ticket for it. :)

opensquirrel/circuit.py
rares1609: Indeed, this makes sense, I'll set it in `__init__`. Also, regarding your second general comment, Chris recommended (in a comment on the [Jira issue](https://qutech-sd.atlassian.net/jira/software/c/projects/CQT/boards/18?selectedIssue=CQT-429)) that it would be better for the initial mapping to be the identity mapping rather than `None`, and frankly it makes sense I suppose.

opensquirrel/circuit.py
rares1609: I will try this

rares1609: Indeed this works, and it makes all the sense in the world to have it like this :)

davidbakker123 (approved):
Looks good! :D

PR #684, opened 2026-05-21
[CQT-342] Implement two qubit canonical decomposition into CZ gates

elenbaasc:


opensquirrel/ir/semantics/canonical_gate.py
```diff
@@ -133,6 +133,32 @@ def is_identity(self) -> bool:
...
+    def __eq__(self, other: Any) -> bool:
+        """Checks if two CanonicalGateSemantic instances are equal.
+
+        Returns:
+            True if both have the same axis and rotations, False otherwise.
+
+        """
+        if not isinstance(other, CanonicalGateSemantic):
+            return False
+        if not np.allclose(self.axis, other.axis, atol=1e-9):
```
davidbakker123: The `BaseAxis` class already has an implementation of this, so you should be able to do `self.axis == other.axis`.

opensquirrel/ir/semantics/canonical_gate.py
```diff
@@ -133,6 +133,32 @@ def is_identity(self) -> bool:
...
+        if self.rotations is None or other.rotations is None:
+            return False
+        return len(self.rotations) == len(other.rotations) and all(
+            r1 == r2 for r1, r2 in zip(self.rotations, other.rotations, strict=False)
+        )
+
+    def __hash__(self) -> int:
+        """Returns hash for use in sets and dicts.
+
+        Note: Returns a fixed hash since CanonicalGateSemantic contains mutable rotations list.
```
davidbakker123: Doesn't this default the purpose of the hash, the fact that `CanonicalGateSemantic` is actually mutable. Doesn't this create issues when it is used in as set/dict?

opensquirrel/passes/decomposer/can2cz_decomposer.py
```diff
@@ -0,0 +1,124 @@
...
+            Decomposition of the original gate into a sequence of gates.
+
+        """
+        if not isinstance(instruction, TwoQubitGate):
+            return [instruction]
+
+        gate = instruction
+        q0, q1 = gate.qubit_operands
+
+        if gate == CZ(q0, q1):
```
davidbakker123: I guess you should then also check if `gate == CZ(q1, q0)`?

opensquirrel/passes/decomposer/can2cz_decomposer.py
```diff
@@ -0,0 +1,124 @@
...
+            return [gate]
+
+        if gate == CNOT(q0, q1):
```
davidbakker123, line 45: I guess you should then also check if `gate == CNOT(q1, q0)`?

opensquirrel/passes/decomposer/can2cz_decomposer.py
```diff
@@ -0,0 +1,124 @@
...
+        gate_axis = gate.canonical.axis
+        gate_rotations = gate.canonical.rotations
+        if gate_rotations is None:
+            gate_rotations = [BlochSphereRotation(axis=(1, 0, 0), angle=0, phase=0)] * 4
+        k1 = SingleQubitGate(q0, gate_rotations[0])
+        k2 = SingleQubitGate(q1, gate_rotations[1])
+        k3 = SingleQubitGate(q0, gate_rotations[2])
+        k4 = SingleQubitGate(q1, gate_rotations[3])
+
+        if np.allclose(gate_axis.value, np.array([0.5, 0, 0])):
```
davidbakker123: Again, you could write this as `gate_axis == CanonicalAxis((0.5, 0.0, 0.0))`

opensquirrel/passes/decomposer/can2cz_decomposer.py
```diff
@@ -0,0 +1,124 @@
...
+                S(q1),
+                H(q1),
+                Ry(q1, -pi / 2),
+                CZ(q0, q1),
+                Ry(q1, pi / 2),
+                H(q0),
+                k3,
+                k4,
+            ]
+        if np.isclose(gate_axis.value[2], 0):
```
davidbakker123: The `BaseAxis` is also indexable, so you can do `gate_axis[2]` directly.

opensquirrel/ir/semantics/canonical_gate.py
elenbaasc: Indeed :)

opensquirrel/ir/semantics/canonical_gate.py
elenbaasc: You're right, this is could lead to issues. I'm not too sure anymore, why I implemented it like this, but it seems unnecessary (and potentially harmful), so I removed it. Thanks for spotting it!

opensquirrel/passes/decomposer/can2cz_decomposer.py
elenbaasc: True, because indeed the CZ gate is symmetrical.

opensquirrel/passes/decomposer/can2cz_decomposer.py
elenbaasc: That could be done, but `CNOT(q0, q1) =/= CNOT(q1, q0)` so we would need to add an extra if-statement. Also note that `CNOT(q1, q0)` is not a standard gate (given the definition of `CNOT(q0, q1)`).

opensquirrel/passes/decomposer/can2cz_decomposer.py
elenbaasc: I remember it complaining about that, but somehow it is fine with it now... anyway, I fixed it

opensquirrel/passes/decomposer/can2cz_decomposer.py
davidbakker123: Aha, you are right. `q0` should always be the control qubit, and `q1` always the target.

davidbakker123 (approved):

PR #687, opened 2026-06-03
[CQT-452] Update OpenSquirrel

elenbaasc:
- Two-qubit gates added to OpenSquirrel default gates:
  - `CV`
  - `CY`
  - `DCNOT`
  - `ECR`
  - `ISWAP`
  - `InvSqrtSWAP`
  - `M`
  - `MS`
  - `SqrtISWAP`
  - `SqrtSWAP`
- Added new gates to existing tests
- Added entry to the Changelog

- This can be reviewed after libQASM has been upgraded.


opensquirrel/ir/default_gates/two_qubit_gates.py
```diff
@@ -102,3 +69,243 @@ def __init__(self, control_qubit: QubitLike, target_qubit: QubitLike, k: Support
...
+            gate_semantic=MatrixGateSemantic(
+                matrix=np.array(
+                    [
+                        [1, 0, 0, 0],
+                        [0, 1, 0, 0],
+                        [0, 0, (1 + 1j) / 2, (1 - 1j) / 2],
+                        [0, 0, (1 - 1j) / 2, (1 + 1j) / 2],
+                    ],
+                ),
+            ),
```
davidbakker123: Why define it with a `MatrixGateSemantic` and not a `ControlledGateSemantic`?

opensquirrel/passes/decomposer/can2cz_decomposer.py
```diff
@@ -40,7 +40,9 @@ def decompose(self, instruction: Gate) -> list[Gate]:
         q0, q1 = gate.qubit_operands
 
         if gate == CZ(q0, q1) or gate == CZ(q1, q0):
-            return [gate]
+            if isinstance(gate, CZ):
```
davidbakker123, line 43: Why do you need this extra `isinstance` check?

tests/integration/test_spin_2_plus.py
davidbakker123, line 1: I see you circumvented the issue with different specific decomposition's by comparing the before/after circuit matrix, nice :)
Two questions:
 - Why did you remove lines 67 and 68?
 - By comparing the matrices, you are not checking whether the control instructions are properly preserved. Is that an issue?

tests/integration/test_tuna_5.py
```diff
@@ -112,81 +115,59 @@ def test_complete_circuit(self, qs_is_installed: bool, data: DataType) -> None:
...
             ]
+
+            cz_operation_re = re.compile(r"^CZ \(q\[(\d+)\], q\[(\d+)\]\)$")
+            cz_pairs = []
+            for operation in operations:
+                cz_match = cz_operation_re.match(operation)
+                if cz_match:
+                    cz_pairs.append((int(cz_match.group(1)), int(cz_match.group(2))))
+
+            assert Counter(cz_pairs) == Counter({(0, 2): 1, (2, 1): 3, (1, 2): 1})
```
davidbakker123:
```suggestion
            cz_pairs = [m.groups() for m in map(cz_operation_re.match, operations) if m]
            assert Counter(cz_pairs) == Counter({('0', '2'): 1, ('2', '1'): 3, ('1', '2'): 1})
```

davidbakker123: Sorry, I thought this was a neat trick, I like `map` :)

tests/passes/decomposer/test_can2cz_decomposer.py
```diff
@@ -51,6 +51,20 @@ def test_decomposes_CZ(decomposer: Can2CZDecomposer, gate: Gate, expected_result
...
+    [
+        CR(0, 1, pi),
+        CRk(0, 1, 1),
+    ],
+    ids=["CR_pi", "CRk_1"],
+)
+def test_decomposes_cr_crk_to_cz(decomposer: Can2CZDecomposer, gate: Gate) -> None:
+    decomposed_gate = decomposer.decompose(gate)
+    check_gate_decomposition(gate, decomposed_gate)
+    assert decomposed_gate == [CZ(0, 1)]
```
davidbakker123, line 66: In the `__eq__`-method you check whether the matrices are equal, so are you really checking that the `decomposed gate` _is_ `CZ`?

opensquirrel/passes/decomposer/can2cz_decomposer.py
elenbaasc: If you have, e.g., a `CR(pi)` gate, the conditions will be satisfied and then you should only return the `gate` if it is in fact a `CZ` gate, because otherwise the `CR(pi)` gate is returned. We can, of course, remove the check and just always return `CZ(q0, q1)`.

tests/integration/test_tuna_5.py
elenbaasc: Don't apologize, I like `map` too 😀! 

Edit: the type checker has issues with passing (overloaded) `re.match` to `map`, so I'm using a generator instead. But still a one-liner 😄

tests/passes/decomposer/test_can2cz_decomposer.py
elenbaasc: Good catch, I'll add an `isinstance` check to see if it is a `CZ` specifically.

tests/integration/test_spin_2_plus.py
elenbaasc: - Lines 67 and 68 can indeed be restored. I had issues with those lines before and tested them somewhere else, but forgot to restore them here.
- Not directly, indeed, but the semantics of the circuit should ultimately match, the validator will check if the interactions are in the form of only CZ gates, and whether the control instructions are preserved is tested specifically somewhere else, e.g. `test_can2cz_decomposer.py`.

davidbakker123 (approved):
All good to go! :)

PR #698, opened 2026-07-01
[CQT-459] Add __contains__ method to the CircuitBuilder

rares1609:


opensquirrel/circuit_builder.py
```diff
@@ -80,6 +80,13 @@ def __getattr__(self, attr: str) -> Any:
         # Default behaviour
         return self.__getattribute__(attr)
 
+    def __contains__(self, item: object) -> bool:
+        if isinstance(item, QubitRegister):
+            return item.name in self.register_manager._qubit_registry  # noqa: SLF001
+        if isinstance(item, BitRegister):
+            return item.name in self.register_manager._bit_registry  # noqa: SLF001
```
elenbaasc: Add the `qubit_registery` and `bit_registry` (getter) properties to the `RegisterManager` and remove the directives here:

```suggestion
            return item.name in self.register_manager.qubit_registry
        if isinstance(item, BitRegister):
            return item.name in self.register_manager.bit_registry
```

elenbaasc (commented):
Looks good!

Just a simple change required to not access private attributes from the `RegisterManager`.

opensquirrel/circuit_builder.py
rares1609: Thanks!

After over a year and a half, I should get over just placing lazy `#noqa`s 😆

elenbaasc (approved):
All good! Thanks 😄

PR #703, opened 2026-07-21
[CQT-461] Make CircuitAnalyzer metrics configurable

SoufiTNO:
- Add metrics and exclude_metrics parameters to (de)select metrics
- Add configurable per-metric timeout parameter
- Add available_metrics() class method
- Add documentation for the CircuitAnalyzer under a new Analysis section in the compilation passes docs
- Add tests for metric selection and timeout
- Update CHANGELOG


opensquirrel/passes/analyzer/circuit_analyzer.py
```diff
@@ -11,6 +14,8 @@
 from opensquirrel.passes.analyzer.general_analyzer import Analyzer
 
 if TYPE_CHECKING:
+    from collections.abc import Iterable
+
```
elenbaasc, line 17:
```suggestion
```

docs/compilation-passes/analysis/index.md
```diff
@@ -0,0 +1,7 @@
+Analyzer passe in OpenSquirrel is used to compute statistics that describe the structure of a quantum circuit.
```
elenbaasc:
```suggestion
Analyzer passes are used in OpenSquirrel to compute statistics that describe the structure of a quantum circuit.
```

docs/compilation-passes/analysis/index.md
```diff
@@ -0,0 +1,7 @@
...
+Unlike most other passes, an analyzer does not modify the circuit: it inspects it and returns a set of metrics.
+These metrics can be used, for instance, to characterise a circuit prior to compilation, to compare circuits, or as input features for machine-learning models that predict compilation performance.
```
elenbaasc:
```suggestion
These metrics can be used, for instance, 
- to characterize a circuit prior to compilation,
- to compare circuits, or
- as input features for machine-learning models that predict compilation performance.
```

docs/compilation-passes/analysis/circuit-analyzer.md
```diff
@@ -0,0 +1,113 @@
+This pass computes a collection of structural metrics that describe a quantum circuit.
+The metrics follow the circuit profiling approach proposed by Bandić _et al._[^1] and are grouped into four categories:
```
elenbaasc:
```suggestion
The metrics follow the circuit profiling approach proposed by Bandić _et al._ (2025)$^1$ and are grouped into four categories:
```

docs/compilation-passes/analysis/circuit-analyzer.md
```diff
@@ -0,0 +1,113 @@
...
+The `timeout` parameter can be combined with `metrics` and `exclude_metrics`.
+
+!!! note "Interrupting a computation"
+
+    The time-out bounds how long the analyzer _waits_ for a metric.
+    The underlying computation runs in a separate thread and cannot be forcibly interrupted,
+    so it may continue in the background until it completes on its own.
+    A positive value is required for `timeout`; a non-positive value raises a `ValueError`.
+
+[^1]:
```
elenbaasc:
```suggestion
$^1$
```

docs/compilation-passes/analysis/circuit-analyzer.md
```diff
@@ -0,0 +1,113 @@
+This pass computes a collection of structural metrics that describe a quantum circuit.
+The metrics follow the circuit profiling approach proposed by Bandić _et al._[^1] and are grouped into four categories:
+
+- **Size**: number of qubits, number of gates, number of two-qubit gates, two-qubit gate percentage, and circuit depth.
+- **Interaction graph (IG)**: metrics derived from the qubit interaction graph, where nodes are qubits and edges are two-qubit gates (_e.g._, diameter, average degree, clustering coefficient).
+- **Gate dependency graph (GDG)**: metrics derived from the directed acyclic graph of gate-to-gate dependencies on shared qubits (_e.g._, critical path length).
+- **Density**: parallelisation-related metrics (density score and idling score).
```
elenbaasc:
```suggestion
1. **Size**: number of qubits, number of gates, number of two-qubit gates, two-qubit gate percentage, and circuit depth.
2. **Interaction graph (IG)**: metrics derived from the qubit interaction graph, where nodes are qubits and edges are two-qubit gates (_e.g._, diameter, average degree, clustering coefficient).
3. **Gate dependency graph (GDG)**: metrics derived from the directed acyclic graph (DAG) of gate-to-gate dependencies on shared qubits (_e.g._, critical path length).
4. **Density**: parallelisation-related metrics (density score and idling score).
```

docs/compilation-passes/analysis/circuit-analyzer.md
```diff
@@ -0,0 +1,113 @@
...
+    ```
+
+## Selecting metrics
+
+Some metrics can be expensive to compute on large circuits.
+For instance, determining the average shortest path in the interaction graph can be costly.
+To avoid computing metrics that are not needed, the analyzer accepts two parameters:
+
+- `metrics`: an iterable of metric names to compute. When omitted, all metrics are computed.
+- `exclude_metrics`: an iterable of metric names to leave out. It is applied after `metrics`.
```
elenbaasc: Every sentence is on a new line.
```suggestion
- `metrics`: an iterable of metric names to compute.
When omitted, all metrics are computed.
- `exclude_metrics`: an iterable of metric names to leave out.
It is applied after `metrics`.
```

docs/compilation-passes/analysis/circuit-analyzer.md
```diff
@@ -0,0 +1,113 @@
...
+!!! note "Unknown metric names"
+
+    Passing a metric name that does not exist to either `metrics` or `exclude_metrics` raises a `ValueError`
+    listing the available metric names.
+
+## Setting a time-out
+
+A per-metric time-out (in seconds) can be set through the `timeout` parameter.
+Each metric is computed under this time budget; if a metric exceeds it, its value is set to `None`, a `UserWarning`
+is emitted, and the analysis continues with the remaining metrics.
```
elenbaasc:
```suggestion
Each metric is computed under this time budget; if a metric exceeds it, its value is set to `None`, 
a `UserWarning` is emitted, and the analysis continues with the remaining metrics.
```

opensquirrel/passes/analyzer/circuit_analyzer.py
```diff
@@ -44,32 +99,62 @@ def analyze(self, circuit: Circuit) -> dict[str, Any]:
...
+        executor = ThreadPoolExecutor(max_workers=1)
+        try:
+            for name in self.metrics:
+                future = executor.submit(self._METRIC_REGISTRY[name], self)
+                try:
+                    metrics[name] = future.result(timeout=self.timeout)
+                except FuturesTimeoutError:
+                    metrics[name] = None
+                    warnings.warn(
+                        f"computation of metric '{name}' exceeded the time-out of {self.timeout} s; "
```
elenbaasc:
```suggestion
                        f"computation of metric '{name}' exceeded the time-out of {self.timeout} s: "
```

opensquirrel/passes/analyzer/circuit_analyzer.py
```diff
@@ -30,10 +35,60 @@ class CircuitAnalyzer(Analyzer):
...
+        self,
+        metrics: Iterable[str] | None = None,
+        exclude_metrics: Iterable[str] | None = None,
+        timeout: float | None = None,
+        **kwargs: Any,
+    ) -> None:
+        super().__init__(**kwargs)
+
+        requested = set(self._METRIC_REGISTRY) if metrics is None else set(metrics)
+        excluded = set(exclude_metrics) if exclude_metrics is not None else set()
```
elenbaasc:
```suggestion
        requested = set(self._METRIC_REGISTRY) if not metrics else set(metrics)
        excluded = set(exclude_metrics) if exclude_metrics else set()
```

opensquirrel/passes/analyzer/circuit_analyzer.py
```diff
@@ -249,31 +354,57 @@ def _critical_path_membership_fraction(
...
+        "ig_clustering_coefficient": _metric_ig_clustering_coefficient,
+        # Gate dependency graph
+        "gdg_critical_path_length": _metric_gdg_critical_path_length,
+        "gdg_path_length_mean": _metric_gdg_path_length_mean,
+        "gdg_path_length_std": _metric_gdg_path_length_std,
+        "gdg_pct_gates_in_critical_path": _metric_gdg_pct_gates_in_critical_path,
+        # Density
+        "density_score": _metric_density_score,
+        "idling_score": _metric_idling_score,
+    }
```
elenbaasc, line 409: Any particular reason why this (class variable) is not defined before the `__init__`?

opensquirrel/passes/analyzer/circuit_analyzer.py
elenbaasc: Missed this one last time. Just to double check, doesn't `pstdev` already return the standard deviation (i.e. not the variance), so why is the square-root returned?

opensquirrel/passes/analyzer/circuit_analyzer.py
```diff
@@ -44,32 +99,62 @@ def analyze(self, circuit: Circuit) -> dict[str, Any]:
...
-        metrics.update(self._gate_dependency_graph_metrics())
-        metrics.update(self._density_metrics())
+        if self.timeout is None:
+            return {name: self._METRIC_REGISTRY[name](self) for name in self.metrics}
+        return self._analyze_with_timeout()
 
+    def _analyze_with_timeout(self) -> dict[str, Any]:
+        """Compute each selected metric in a worker thread, bounded by the configured time-out."""
+        metrics: dict[str, Any] = {}
+        executor = ThreadPoolExecutor(max_workers=1)
```
elenbaasc, line 110: Python cannot kill a running thread. So if the timeout error is raised the calculation will/might still run in the background. You could consider using `multiprocessing`, to replace thread pool timeout with terminating a process.

elenbaasc (commented):
Very nice implementation, thanks for the effort!

My main concern is about using thread pool timeout, as the calculation will/might continue in the background, instead of being killed.

opensquirrel/passes/analyzer/circuit_analyzer.py
SoufiTNO: The registry points to the _metric_* methods so it has to come after them in the class body, if I move it above __init__, those methods dont exist yet and it errors

opensquirrel/passes/analyzer/circuit_analyzer.py
SoufiTNO: My bad, I confused pstdev with another function. Thank you Chris :)
I will fix it to return the correct value

opensquirrel/passes/analyzer/circuit_analyzer.py
SoufiTNO: You re right that the thread cant be killed, so a slow metric may keep running in the background. I went with threads because they re lighter , multiprocessing would be heavier to set up and slower to start for each metric. Since the timeout is just a safety net for the occasional slow metric I thought threads were a fair tradeoff. But I am happy to switch to multiprocessing if you prefer the calculation to actually stop

elenbaasc: Thanks for the explanation. Let's keep it like this for now.

I have made a separate ticket, if we would want to pick this up in the future.

elenbaasc (approved):
Good to go! Thanks for resolving the comments 🙂

PR #709, opened 2026-08-24
[CQT-466] Increase test coverage

rares1609:


tests/ir/semantics/test_canonical_gate_semantic.py
```diff
@@ -61,3 +90,34 @@ def test_invalid_number_of_rotations(self) -> None:
...
+
+    def test_eq_without_rotations(self, semantic: CanonicalGateSemantic) -> None:
+        assert semantic == CanonicalGateSemantic((0, 0, 0))
+
+    def test_eq_with_rotations(self, semantic_with_rotations: CanonicalGateSemantic) -> None:
+        assert semantic_with_rotations == CanonicalGateSemantic((0.25, 0.25, 0.25), semantic_with_rotations.rotations)
+
+    def test_eq_only_one_has_rotations(self, semantic_with_rotations: CanonicalGateSemantic) -> None:
+        assert semantic_with_rotations != CanonicalGateSemantic((0.25, 0.25, 0.25))
+        assert CanonicalGateSemantic((0.25, 0.25, 0.25)) != semantic_with_rotations
```
elenbaasc: I don't understand the point of these asserts?

tests/ir/test_ir.py
```diff
@@ -100,3 +103,15 @@ def test_cnot_inequality(self) -> None:
 
     def test_hash_difference_bit_qubit(self) -> None:
         assert hash(Qubit(1)) != hash(Bit(1))
+
+    def test_eq_different_type(self) -> None:
+        assert IR() != "not an IR"
```
elenbaasc: This is too obvious to be a test.

tests/ir/test_ir.py
```diff
@@ -100,3 +103,15 @@ def test_cnot_inequality(self) -> None:
...
+
+    def test_repr(self) -> None:
+        assert repr(IR()) == "IR: []"
```
elenbaasc: Would be more interesting (better test), to add some statements. This will force one to make sure to define a `__repr__` method for each statement.

tests/ir/semantics/test_canonical_gate_semantic.py
rares1609: They are meant to check that `__eq__ ` is symmetric.

The guard in `canonical_gate.py` is `self.rotations is None or other.rotations is None`. I thought that whichever object sits on the left of the `!=` is the one that becomes `self`, so swapping the order would mean that each assert covers a different half of that `or`.

Maybe I should clearly rename it to `test_eq_is_symmetric_when_only_one_has_rotations`? Or do you think the test is just redundant and I should get rid of it?

tests/ir/test_ir.py
rares1609: You are right, it is utterly redundant. Removed it.

tests/ir/test_ir.py
rares1609: Thank you for the suggestion Chris! Indeed, this does make the test a lot more meaningful. I have applied your suggestion, for `Barrier`, `Init` and `Measure`. While I was at it, I noticed that `SingleQubitGate` and `AsmDeclaration` don't implement `__repr__`, and neither do any of their bases (i.e. `Unitary`, `Gate`, `Statement` etc.), so they fall back to `object.__repr__`.

Should I go ahead and add the `__repr__` method to e.g. `SingleQubitGate` and `AsmDeclaration`, and then add the relevant statements to `test_repr`?

tests/ir/semantics/test_canonical_gate_semantic.py
elenbaasc: Indeed the l.h.s. is `self` and the r.h.s. is the `other`. Nevertheless, I would say that this test is redundant, because there is no asymmetry in the current implementation and it would be rather strange to introduce one later on.

tests/ir/test_ir.py
elenbaasc: Yes, it would be good to add `__repr__`s to the ones that don't have one. Fine to do it in this PR. The AsmDeclatation, might be and interesting one, because I think we should not include the content in the representation. But have a look if you think it should.

tests/ir/semantics/test_canonical_gate_semantic.py
rares1609: Removed it

tests/ir/test_ir.py
rares1609: Implemented `__repr__` for `SingleQubitGate` and `AsmDeclaration`.

opensquirrel/ir/single_qubit_gate.py
```diff
@@ -108,3 +108,7 @@ def __mul__(self, other: SingleQubitGate) -> SingleQubitGate:
             msg = "cannot merge two single-qubit gates on different qubits"
             raise ValueError(msg)
         return SingleQubitGate(self.qubit, self.bsr * other.bsr)
+
+    def __repr__(self) -> str:
+        gate_semantic = self._bsr if self._bsr is not None else self._matrix
```
elenbaasc: Because of the `bsr` property, it should always be possible to get the Bloch sphere representation. If it is none, it will be generated from the matrix, so the `__repr__` will consistently provide the bsr.

opensquirrel/ir/statement.py
```diff
@@ -30,6 +30,9 @@ def accept(self, visitor: IRVisitor) -> Any:
         visitor.visit_statement(self)
         return visitor.visit_asm_declaration(self)
 
+    def __repr__(self) -> str:
+        return f"AsmDeclaration(backend_name={self.backend_name})"
```
elenbaasc: Probably good to add just a bit extra info/meta data:
```suggestion
        return (
            f"AsmDeclaration(backend_name={self.backend_name}, "
            f"backend_code_length={len(str(self.backend_code))})"
        )
```

elenbaasc (approved):
Great, thanks for making the final adjustments!

PR #710, opened 2026-08-28
[CQT-470] Fix documentation of the Measure decomposer

elenbaasc:


docs/compilation-passes/decomposition/measure-decomposer.md
```diff
@@ -3,9 +3,11 @@ The measure decomposer (`MeasureDecomposer`) decomposes measurements along an ar
 ## Theory
 
 For a measurement axis defined by the unit vector
+
 $$\hat{n} = (\sin\theta\cos\phi, \sin\theta\sin\phi, \cos\theta)$$
 
-where $\theta = \arccos(n_z)$ and $\phi = \arctan2(n_y, n_x)$, we construct the unitary transformation:
```
davidbakker123, line 8: The convention [atan2](https://en.wikipedia.org/wiki/Atan2) seems to be more common than with a subscript 2.

PR #711, opened 2026-09-08
[CQT-467] Add three-qubit gates CCX and CSWAP

SoufiTNO:
I added ThreeQubitGate with CCX (alias CCNOT) and CSWAP, a ThreeQubitGateDecomposer into CZ and single-qubit gates, and cQASM v1 export of the Toffoli gate.

The CCX decomposition uses 6 CZ gates, which [Shende and Markov (2008)](https://arxiv.org/abs/0803.2316) prove is minimal. CSWAP uses 8, but that's an upper bound only: the Fredkin CNOT-cost is still open. Both preserve global phase exactly.


elenbaasc (commented):
Very nice work here, great implementation!

I have no comments on the changes made 🙂!

What is still missing though, is the additions to the **User documentation**, i.e., a page on the ThreeQubitGateDecomposer, adding the three qubit gates to the list of gates that the CircuitBuilder accepts, and it would be good to add the toffoli gate to (some of) the examples of the cQASM v1 exporter.

docs/compilation-passes/decomposition/predefined-decomposers.md
```diff
@@ -1,11 +1,12 @@
 We refer to a decomposition as being _predefined_ when each instance of a particular gate in the circuit,
 _e.g._, the CNOT gate, is replaced with a fixed gate or list of gates.
 
-Three predefined decomposers are available in OpenSquirrel:
+Four predefined decomposers are available in OpenSquirrel:
 
 - CNOT to CZ decomposer
 - SWAP to CNOT decomposer
 - SWAP to CZ decomposer
+- Three-qubit gate decomposer
```
elenbaasc, line 9: Good call placing it under the predefined decomposers! I agree that it should be here, indeed.

elenbaasc (approved):
Looks good to me! Nice addition of the documentation (and also thanks for spotting the missing tow-qubit gates and mistakes in the docs of the cQASM v1 exporter).

PR #712, opened 2026-09-14
[CQT-475] Update user documentation with the newly added 2-qubit gates

SoufiTNO:
Adds the missing two-qubit gates to the instruction list, updates the SWAP decomposition note now that Can2CZDecomposer can decompose SWAP, and fixes the class name and class-vs-instance typos in the cQASM v1 exporter examples.


elenbaasc (approved):
Very nice, thanks for catching this and fixing it!

